PR #11, opened 2018-06-19
Message: fix missing release of sd_bus_message

lukasdurfina:
The fix for a memory leak, which occurs with an usage of sdbus::Variant.


sangelovic:
Indeed the unref was missing there. Thank you.

PR #12, opened 2018-06-29
Introduce support for asynchronous server-side methods

sangelovic:
Please squash and merge.

FYI @Oberonc


lukasdurfina:
Very nice! Thank you

PR #19, opened 2018-08-24
stub-generator: fixed issue when <arg> in <method> does not have "name"

lukasdurfina:
E.g https://github.com/connectivity/connman-gnome/blob/master/common/connman-agent.xml
is missing "name" for method arguments.

This patch ensures that the valid code is generated for such a case.

### DIFF
--- a/stub-generator/BaseGenerator.cpp
+++ b/stub-generator/BaseGenerator.cpp
@@ -108,12 +108,21 @@ std::tuple<std::string, std::string, std::string> BaseGenerator::argsToNamesAndT
 {
     std::ostringstream argSS, argTypeSS, typeSS;
 
-    bool firstArg{true};
-    for (const auto& arg : args)
+    for (auto i = 0; i < args.size(); ++i)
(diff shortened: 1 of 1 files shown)


stub-generator/BaseGenerator.cpp
```diff
@@ -108,12 +108,21 @@ std::tuple<std::string, std::string, std::string> BaseGenerator::argsToNamesAndT
 {
     std::ostringstream argSS, argTypeSS, typeSS;
 
-    bool firstArg{true};
-    for (const auto& arg : args)
+    for (auto i = 0; i < args.size(); ++i)
     {
-        if (firstArg) firstArg = false; else { argSS << ", "; argTypeSS << ", "; typeSS << ", "; }
+        auto arg = args.at(i);
```
sangelovic: Would generally be safer to use decltype(auto), which yields reference, and it retains behavior of previous range based loop. But here it's probably ok to copy Node object.

PR #22, opened 2018-09-16
add support for type SD_BUS_TYPE_UNIX_FD

lrusak:
I doubt this is correct but I just wanted to PR it because it fixes my runtime issue.

### DIFF
--- a/src/Message.cpp
+++ b/src/Message.cpp
@@ -227,6 +227,12 @@ Message& Message::operator>>(int32_t& item)
     if (r == 0)
         ok_ = false;
 
+    if (r < 0) {
+        r = sd_bus_message_read_basic((sd_bus_message*)msg_, SD_BUS_TYPE_UNIX_FD, &item);
+        if (r == 0)
+            ok_ = false;
+    }
+
     SDBUS_THROW_ERROR_IF(r < 0, "Failed to deserialize a int32_t value", -r);
 
     return *this;


lukasdurfina:
Hi,
I think your patch is not fixing the issue. According to https://www.freedesktop.org/wiki/Software/systemd/logind/ the return type of `Inhibit` is `h`. And according to https://dbus.freedesktop.org/doc/dbus-specification.html it is **Unsigned 32-bit integer representing an index into an out-of-band array of file descriptors, transferred via some platform-specific mechanism (mnemonic: h for handle)**

I guess the correct type for `fd` should fix it. Try `uint32_t` and let me know.

xyzzy42:
The type should be `int`.  Are you sure your read code gets to the case of trying the unix fd?  Maybe there is no error when using `SD_BUS_TYPE_INT32`.

You can see where this is handled in `sd_bus_message_read_basic()`,

```
                        case SD_BUS_TYPE_UNIX_FD: {
                                uint32_t j;

                                j = BUS_MESSAGE_BSWAP32(m, *(uint32_t*) q);
                                if (j >= m->n_fds)
                                        return -EBADMSG;

                                if (p)
                                        *(int*) p = m->fds[j];

```
You can tell from the cast at the end that the pointer passed should be one to an `int`.  Note that `int32_t` and `int` are not necessarily the same.  The conversion from uint32_t to int and dereferencing the fds array (the out of band array of fds mentioned before) is handled by systemd here.

I believe the concept of tying the type of the dbus return value to the C++ type will not work here.  File descriptors and signed integers are the same C++ type, but they are not the same dbus type.

Consider creating a class for file descriptors.  It can nothing more than a struct with a single int field for the fd and default constructors.

lrusak:
So a little more investigation yields
```
  what():  [System.Error.ENXIO] Failed to deserialize a int32_t value (No such device or address)
```

So I am getting back an int32_t (int on my machine)

also, looking at the sdbus test-inhibit.c I see that they `dup(fd)` https://github.com/systemd/systemd/blob/master/src/login/test-inhibit.c#L12

I've updated the PR (I still don't think this is the right solution) but it works now.

This is the test code I'm using
```
#include <errno.h>
#include <string.h>
#include <stdio.h>
#include <unistd.h>
#include <fcntl.h>

#include <sdbus-c++/sdbus-c++.h>

int main()
{
  auto proxy = sdbus::createObjectProxy("org.freedesktop.login1", "/org/freedesktop/login1");

  int32_t fd;
  proxy->callMethod("Inhibit").onInterface("org.freedesktop.login1.Manager").withArguments("sleep", "test-app", "test", "delay").storeResultsTo(fd);

  printf("fd: %d\n", fd);

  if (fd >= 0)
  {
    auto ret = fcntl(fd, F_GETFD);
    if (ret < 0)
    {
      printf("errror: %s (%d)\n", strerror(errno), errno);
      return -1;
    }

    close(fd);
  }
}
```

lechndo:
Hi guys,

the discussion looks a bit outdated.
Nevertheless I want to show you my approach to solve this issue.
It is about using a wrapping struct for FDs as suggested by @xyzzy42.
For sure this solution is not yet perfect but may be a good start for a proper fix.

[unixFdSupport.zip](https://github.com/Kistler-Group/sdbus-cpp/files/3246207/unixFdSupport.zip)

lrusak:
> Hi guys,
> 
> the discussion looks a bit outdated.
> Nevertheless I want to show you my approach to solve this issue.
> It is about using a wrapping struct for FDs as suggested by @xyzzy42.
> For sure this solution is not yet perfect but may be a good start for a proper fix.
> 
> [unixFdSupport.zip](https://github.com/Kistler-Group/sdbus-cpp/files/3246207/unixFdSupport.zip)

That looks pretty good to me! the only suggestion I would make is that `0` is a valid fd so it so the default should be `-1`

lechndo:
> That looks pretty good to me! the only suggestion I would make is that `0` is a valid fd so it so the default should be `-1`

Thx for pointing that out. Seems I was a bit too much focused in the DBus definition which  mentions to FDs as  
> Unsigned 32-bit integer
and totally forgot about the basic fact that negative values are used to signal errors or invalid FDs.
Hope the following will work:

[unixFdSupport.zip](https://github.com/Kistler-Group/sdbus-cpp/files/3250856/unixFdSupport.zip)

xyzzy42:
If you're targeting C++14, you can write the class as:
```
struct UnixFD {
    int fd = -1;
};
```
Same effect as supplying a default value in the constructor.  Aggregate initialization allows construction from a single int.  Structured bindings would allow `int [ fd ] = ReturnsUnixFDstruct();`.
Or adding an `operator int() { return fd; }` would allow using a UnixFD where a normal int fd would be used without conversion.

lrusak:
@lechndo Please PR it. Then this PR can be closed

sangelovic:
Hi guys, thank you all for your contributions.

@lechndo Will you please issue a pull request for your patch?

A couple of notes to the patch:
* Please follow @xyzzy42 's suggestion on `UnixFD` struct initialization, and please add the conversion operator as mentioned there...
* Please change tabs to 4 spaces (sdbus-c++ uses spaces)
* Can you also extend unit tests (update `TEST(AMessage, CanCarryAComplexType)` in Message_test.cpp and extend TypeTraits_test.cpp) and integration tests as well? But I can do that myself as well after accepting the PR since I'm familiar with them.

Thanks!

lechndo:
Hey guys,

I'm a bit busy at the moment. Got some other urgent tasks to do...
I hope I can finish the other urgent issues and switch back to my sdbus-c++ evaluation task and then I'll prepare the PR ASAP...

Will also have a look on the unit test stuff and try to add some tests for UnixFDs.

PS:
I again came across a feature we would need which seems to be missing...
We would need a convenient way for sending "property changed" signals... 
So maybe this might be another thing for a contribution... If my bosses will give me the permission to continue...

sangelovic:
Hi, maybe I will have time to introduce UnixFD myself. Would you mind? I plan to release v0.7.1 very soon and PR yocto recipes for that version to open-embedded, and it would be good to have UnixFd support in that version.

PS: Standard D-Bus interfaces (that is, including `org.freedesktop.DBus.Properties` that defines `PropertyChanged` signal) are supported. It's one of the most recent features. See https://github.com/Kistler-Group/sdbus-cpp/blob/master/docs/using-sdbus-c++.md#standard-d-bus-interfaces

lechndo:
> Hi, maybe I will have time to introduce UnixFD myself. Would you mind? 

No. If you have time please go ahead...

> PS: Standard D-Bus interfaces (that is, including `org.freedesktop.DBus.Properties` that defines `PropertyChanged` signal) are supported. It's one of the most recent features. See https://github.com/Kistler-Group/sdbus-cpp/blob/master/docs/using-sdbus-c++.md#standard-d-bus-interfaces

Nice... Seems I've missed the latest added features. Thx for the hint.

mdavis777:
@sangelovic 
Its not much but I happen to already have a Yocto recipe for 0.6.0 you are free to use and build off of.


```
SUMMARY = "D-Bus IPC C++ binding library built on top of sd-bus, a D-Bus C library by systemd"
LICENSE = "LGPLv2.1"
LIC_FILES_CHKSUM = "file://COPYING;md5=1803fa9c2c3ce8cb06b4861d75310742 "

SRC_URI = "https://github.com/Kistler-Group/${BPN}/archive/v${PV}.tar.gz"

SRC_URI[md5sum] = "813b6fd423ae68675cf29231baadc067"
SRC_URI[sha256sum] = "a1084781090842b158ede14c4237b0dd8d9612a41fead835a3aefb099498970b"

DEPENDS = "systemd"
RDEPENDS_${PN} = "systemd"

inherit cmake

EXTRA_OECMAKE += " -DENABLE_TESTS=OFF "
```

sangelovic:
Thanks @mdavis777 . I've started from a similar recipe, too. I've extended it to support ptest and non-systemd environments as well -- the recipe detects whether libsystemd feature is on or off, in the former case it depends on existing libsystemd (like yours does), in the latter case it gets in all dependencies and builds libsystemd as invisible integral part of sdbus-c++, without any run time dep on systemd.

sangelovic:
Closed by #61

PR #27, opened 2018-12-19
Add ability to declare property behavior on PropertyChanged signal

110yd:
Systemd library has ability to declare behavior on properties with annotations like:
`<annotation name="org.freedesktop.DBus.Property.EmitsChangedSignal" value="false|true|const"/>`
This patch adds ability to explicitly specify what behavior do we declare on our property. New functionality related only to object introspection, user still has to emit `org.freedesktop.DBus.Property.PropertiesChanged` signal on his/her own.


sangelovic:
Thank you for your contribution. I'm now going to add support for some more common annotations (Method.NoReply - in progress, Deprecated - tbd). Happy New Year.

PR #31, opened 2019-01-05
Remove warnings-related compiler options from CMakeLists

### LINKED ISSUE #28: 0.3.3 release doesn't build out of the box by unittests build
master is also affected by this issue:
```
sdbus-cpp/test/unittests/Types_test.cpp: In member function ‘virtual void AVariant_CanBeConstructedFromAComplexValue_Test::TestBody()’:
sdbus-cpp/test/unittests/Types_test.cpp:69:35: error: unnecessary parentheses in declaration of ‘value’ [-Werror=parentheses]
     ASSERT_NO_THROW(sdbus::Variant(value));
                                   ^
```

Possible fix there: merhalak@93248b80a2ea1f418f88e545fe933532c690a515

sangelovic:
These options:

* shall rather come from the user instead of being hard-coded in the project build system,
* if kept hard-coded, they may cause unnecessary compilation errors, like in #28 , due to new (and not always useful, cf. discussion at https://github.com/boostorg/mpl/issues/31#issuecomment-368998776) warnings introduced by new versions of compilers.

Fixes #28 , as the warning (turned to error) there is issued by gcc 8 on a completely valid code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,7 +65,6 @@ set(SDBUSCPP_SRCS ${SDBUSCPP_CPP_SRCS} ${SDBUSCPP_HDR_SRCS} ${SDBUSCPP_PUBLIC_HD
 #-------------------------------
 
 set(CMAKE_CXX_STANDARD 17)
-add_compile_options(-W -Wextra -Wall -Werror -pedantic)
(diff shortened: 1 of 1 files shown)


lukasdurfina:
That is a reasonable change. It is being built by many compilers and it is hard to watch for all new warnings.

PR #35, opened 2019-02-04
Add basic 'Connection' unit-tests

ardazishvili:
Test creation of Connection class and basic requests.


ardazishvili:
Found that I occasionally duplicated integration test of Connection class

PR #42, opened 2019-03-15
Introduce support for asynchronous D-Bus method calls on the client side

sangelovic:


sangelovic:
@lukasdurfina I'm merging this now, but let feel free to comment retroactively if you have any comments.

PR #47, opened 2019-03-25
Re-design sdbus-c++ approach to connections

sangelovic:
Changes towards scalability in sdbus-c++ Connections. Simplified usage of connections in ObjectProxy -- no need for two connections. Each connection now should have an event loop thread associated with it, so we don't have a possible inactive connection from the dbus-daemon perspective. Connections are now thread-safe and can now be shared by multiple ObjectProxies -- access is serialized. Async method calls now work flawlessly.

FYI @lukasdurfina


sangelovic:
Tutorial will be updated separately in the next days.

@lukasdurfina Merging again just now, please post comments even post-mortem if you encounter anything

PR #52, opened 2019-05-09
Refactor

ardazishvili:
Most of the changes caused by cland-tidy. Another reason is cpp core guidelines from http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines

PR #62, opened 2019-06-10
Redesign inheritance from Message

sangelovic:
So that the design better, more idiomatically expresses the intent


sangelovic:
Fixes part of #52 (thanks to @ardazishvili )

PR #66, opened 2019-06-25
User subscribes to the signal

lujfsd:
in the proxy , users should subscribes the signal by method SetUpSignal....()


sangelovic:
Can you please describe in greater detail what this change is about and what is the motivation behind it? And provide an example? It is not easy to extract that information from the diff especially when tons of code formatting changes are part of the change set as well (please don't mix feature changes with formatting changes, use a separate PR for formatting changes. (and anyway formatting changes in that code are not necessary at all)).

lujfsd:
sorry for formatting changes，
for examplem,The XML as below:
```
<?xml version="1.0" encoding="UTF-8"?>
<node name="/org/sdbuscpp/concatenator">
    <interface name="org.sdbuscpp.Concatenator">
        <method name="concatenate">
            <arg type="ai" name="numbers" direction="in" />
            <arg type="s" name="separator" direction="in" />
            <arg type="s" name="concatenatedString" direction="out" />
        </method>
        <signal name="concatenated">
            <arg type="s" name="concatenatedString" />
        </signal>
    </interface>
</node>
```
before changed，generate  proxy header file like :
```

/*
 * This file was automatically generated by sdbus-c++-xml2cpp; DO NOT EDIT!
 */

#ifndef __sdbuscpp__testProxy_h__proxy__H__
#define __sdbuscpp__testProxy_h__proxy__H__

#include <sdbus-c++/sdbus-c++.h>
#include <string>
#include <tuple>

namespace org {
namespace sdbuscpp {

class Concatenator_proxy
{
public:
    static constexpr const char* INTERFACE_NAME = "org.sdbuscpp.Concatenator";

protected:
    Concatenator_proxy(sdbus::IProxy& proxy)
        : proxy_(proxy)
    {
        proxy_.uponSignal("concatenated").onInterface(INTERFACE_NAME).call([this](const std::string& concatenatedString){ this->onConcatenated(concatenatedString); });
    }

    virtual void onConcatenated(const std::string& concatenatedString) = 0;

public:
    std::string concatenate(const std::vector<int32_t>& numbers, const std::string& separator)
    {
        std::string result;
        proxy_.callMethod("concatenate").onInterface(INTERFACE_NAME).withArguments(numbers, separator).storeResultsTo(result);
        return result;
    }

private:
    sdbus::IProxy& proxy_;
};

}} // namespaces

#endif

```
look at the constructor ,default subscribes all signals,but sometimes I don't need all。
although I can do this by implementing an empty onSignal function ,but dbus-daemon will still send a signal to the all Concatenator_proxy,When there are many Concatenator_proxy, there will be some impact on performance.
after changed, the proxy like:
```

/*
 * This file was automatically generated by sdbus-c++-xml2cpp; DO NOT EDIT!
 */

#ifndef __sdbuscpp__testProxy_h__proxy__H__
#define __sdbuscpp__testProxy_h__proxy__H__

#include <sdbus-c++/sdbus-c++.h>
#include <string>
#include <tuple>

namespace org {
namespace sdbuscpp {

class Concatenator_proxy
{
public:
    static constexpr const char* INTERFACE_NAME = "org.sdbuscpp.Concatenator";

protected:
    Concatenator_proxy(sdbus::IProxy& proxy)
        : proxy_(proxy)
    {
    }

    ~Concatenator_proxy() = default;

    virtual void onConcatenated(const std::string& concatenatedString) = 0;

public:
    void setUpSignalconcatenated()
    {
        proxy_.uponSignal("concatenated").onInterface(INTERFACE_NAME).call([this](const std::string& concatenatedString){ this->onConcatenated(concatenatedString); });
    }

public:
    std::string concatenate(const std::vector<int32_t>& numbers, const std::string& separator)
    {
        std::string result;
        proxy_.callMethod("concatenate").onInterface(INTERFACE_NAME).withArguments(numbers, separator).storeResultsTo(result);
        return result;
    }

private:
    sdbus::IProxy& proxy_;
};

}} // namespaces

#endif

```
i am not subscribes signal in constructor ,subscribes signal by func setUpSignalconcatenated when i needed.

sangelovic:
Thanks for contribution, but sorry, I won't implement that. I think there is no strong reason for that. I don't think it typically is a performance problem when the client gets a signal whereas the signal is not desired. If one measures and truly sees that the distribution of a signal from D-Bus daemon to the destination process is really the performance bottleneck, then one has bigger problems. These questions should be answered first: Am I not misusing signals for something that is performance-critical (maybe sending some captured data or similar)? If not, i.e. there is no overload of the system with signals, then what I am doing wrong that the bottleneck in my system is just one exchange of D-Bus message directly between D-Bus daemon and my service? The entire system is in trouble then.

PR #67, opened 2019-06-27
[WIP][RFC] Add owning types for unix file descriptors

### LINKED ISSUE #63: When to close a passed file descriptor?
I want my server to pass one end of a socket pair to a client as a return value from a method. I need to close the client side after it is sent. I had a look at sdbus::UnixFd and it does not seem to take ownership. How can I know when it is safe to close the file descriptor so only the client has it?

cvuchener:
Rename current UnixFd to UnixFdRef, add UniqueUnixFd as owning unix file
descriptor, and AnyUnixFd to hold any of the previous two.

Change stub generator to use AnyUnixFd for outgoing file descriptor and
UnixFdRef for incoming file descriptors (they are owned by the message).

Fix #63.

--------

This is not ready yet. I sent the PR so the design could be discussed and to get help to fix issues.

I tested passing fds with this patch, it worked for passing fds from client to server, but not for passing them from server to client through methods or properties (but it worked with signals). Any idea what I did wrong?

About the design, I am not sure about the use of unique_ptr for holding file descriptor, it kind of do everything needed, but it feels like abusing to make file descriptor pointers that cannot be dereferenced.

UniqueUnixFd and AnyUnixFd use inheritance instead of aliases so they can be forward declared and have operator int. But I think it would be cleaner if they could be aliases. Conversion operator are not that important : UniqueUnixFd::operator int is the same thing as unique_ptr::get, and AnyUnixFd::operator int should only be used in operator<<, so std::visit could be called directly there.

### DIFF
--- a/include/sdbus-c++/Message.h
+++ b/include/sdbus-c++/Message.h
@@ -44,7 +44,9 @@ namespace sdbus {
     class ObjectPath;
     class Signature;
     template <typename... _ValueTypes> class Struct;
-    struct UnixFd;
+    struct UnixFdRef;
+    struct UniqueUnixFd;
+    struct AnyUnixFd;
     class MethodReply;
     namespace internal {
         class ISdBus;
@@ -92,7 +94,9 @@ namespace sdbus {
         Message& operator<<(const Variant &item);
         Message& operator<<(const ObjectPath &item);
         Message& operator<<(const Signature &item);
-        Message& operator<<(const UnixFd &item);
+        Message& operator<<(const AnyUnixFd &item);
+        Message& operator<<(const UnixFdRef &item);
+        Message& operator<<(const UniqueUnixFd &item);
 
         Message& operator>>(bool& item);
         Message& operator>>(int16_t& item);
@@ -108,7 +112,8 @@ namespace sdbus {
         Message& operator>>(Variant &item);
         Message& operator>>(ObjectPath &item);
         Message& operator>>(Signature &item);
-        Message& operator>>(UnixFd &item);
+        Message& operator>>(UnixFdRef &item);
+        Message& operator>>(UniqueUnixFd &item);
 
         Message& openContainer(const std::string& signature);
         Message& closeContainer();
--- a/include/sdbus-c++/TypeTraits.h
+++ b/include/sdbus-c++/TypeTraits.h
@@ -34,6 +34,7 @@
 #include <cstdint>
 #include <functional>
 #include <tuple>
+#include <variant>
 
 // Forward declarations
 namespace sdbus {
@@ -41,7 +42,9 @@ namespace sdbus {
     template <typename... _ValueTypes> class Struct;
     class ObjectPath;
     class Signature;
-    struct UnixFd;
+    struct UnixFdRef;
+    struct UniqueUnixFd;
+    struct AnyUnixFd;
     class MethodCall;
     class MethodReply;
     class Signal;
@@ -289,7 +292,29 @@ namespace sdbus {
     };
 
     template <>
-    struct signature_of<UnixFd>
+    struct signature_of<AnyUnixFd>
+    {
+        static constexpr bool is_valid = true;
+
+        static const std::string str()
+        {
+            return "h";
+        }
+    };
+
+    template <>
+    struct signature_of<UnixFdRef>
+    {
+        static constexpr bool is_valid = true;
+
+        static const std::string str()
+        {
+            return "h";
+        }
+    };
+
+    template <>
+    struct signature_of<UniqueUnixFd>
     {
         static constexpr bool is_valid = true;
 
--- a/include/sdbus-c++/Types.h
+++ b/include/sdbus-c++/Types.h
@@ -35,6 +35,10 @@
 #include <memory>
 #include <tuple>
 
+extern "C" {
+#include <unistd.h>
+}
+
 namespace sdbus {
 
     /********************************************//**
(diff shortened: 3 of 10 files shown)


lechndo:
Hm... 
I fear I don't understand all the details of your concept but at least the size of the change set implies a not so simple design.

As you said:
> I think it is best to let the user duplicate the file descriptor as needed.

I would suggest to keep it all rather simple using a single UnixFd class as we had it until now and give the UnixFd class the ability to close it's FD on destruction if the user needs this behavior.

This way on reception the user would get a UnixFd object and either might keep it to use it's FD, dup the FD and use the duplicate or if doesn't need the FD at all simply let the passed object die and the FD will be cleaned up automatically.

When sending an FD the approach is quite similar.
Putting the FD into the UnixFd object before sending will hand over the ownership to the Unixd object and it will close it on destruction. So if the user still needs an open FD to the resource it has to dup the FD before or on construction of the UnixFd object (maybe the object could do this itself somehow if required)

What do you think?

cvuchener:
> I would suggest to keep it all rather simple using a single UnixFd class as we had it until now and give the UnixFd class the ability to close it's FD on destruction *if the user needs this behavior.*

I don't think making a optionally owning object would be simpler. Actually the result would not be much different from the variant I am proposing. A few method could be added to make it more usable if it was used everywhere. But I personally prefer when the ownership is clearly shown in the type.

> Putting the FD into the UnixFd object before sending will hand over the ownership to the Unixd object and it will close it on destruction. So if the user still needs an open FD to the resource it has to dup the FD before or on construction of the UnixFd object (maybe the object could do this itself somehow if required)

This would be a very simple solution. It only requires to add a close in operator<<. But this adds more work for the user if they want to keep the file descriptor, and the file descriptor would be duplicated twice. It may be annoying to the lower level API users too.

cvuchener:
I found out the reason for my issue. I did not think of the consequence of the ownership of received file descriptors. It worked with signals because the message object still exists when the callback is called. But when using proxy methods, the return value is already closed as the message was freed, the lifetime must be extended in this case.

Solving that would make the stub generator more complex, it would need to use the three types:
 - AnyUnixFd for adaptor methods/properties return values (for leaving the choice to the user),
 - UniqueUnixFd for proxy methods/properties return values (for extending the lifetime),
 - UnixFdRef for function arguments (because they are not taking ownership, and taking a lvalue reference to a UniqueUnixFd as an argument is not taking ownership anyway, it's just useless).

lechndo:
My suggestion was to always let the `UnixFd` class own the FD and also close it on destruction.
So the ownership is clear.

The only burden for the user is to decide if he wants to keep and use the FD in the future after the lifetime of the `UnixFd` object in which case he has to take the FD and `dup` it.

FD handling is always tricky so the user has to be careful in any case so this burden is not too much I think.

If I didn't miss some special cases this solution should work all scenarios and doesn't require a change to the stub generator.

sangelovic:
Hi guys, I was on holidays so sorry for a bit of delay. Nice to read your discussion, thanks a lot for sharing your thoughts!

So as I see it we can sum up the underlying sd-bus behavior regarding fd like this:

* On sender side, sd-bus duplicates fd that is being passed into the message and stores this duplicated fd, leaving the original, passed-in fd in the ownership of the caller.
* sd-bus message becomes an owner of the (duplicated) fd, and the recipient, when reading out the fd, must be aware of that fact, and either use fd and let it be closed on message destruction, or duplicate it in order to be able to use it even after the message has been freed.

So far so good.

But there are currently two problems in sdbus-c++ that are related to that:

1. Inability to close the fd by the user when returning the fd from a D-Bus method implemented using convenience API (as reported at #63).
2. Inability to get the fd from the reply on the client side because that fd has been closed right after deserialization, but before the user gets it, when the reply message was freed. This again when using convenience API, and thus generated stubs as well.

Clearly -- as we all agree -- we need a form of RAII-style `UnixFd` implementation that assumes ownership of its fd, but also provides certain flexibility upon construction. So such a `UnixFd`:

* Provides default constructor
* Provides constructor overload that takes the fd argument and duplicates it, leaving the responsibility for the passed-in fd up to the user, but assuming ownership of the duplicate fd.
* Provides constructor overload that takes the fd argument as-is and assumes ownership of it (a-ka "move" semantics, by having an additional `adopt_fd_t` parameter, just like e.g. [unique_lock](https://en.cppreference.com/w/cpp/thread/unique_lock/unique_lock) in STL).
* For simplicity, provides default move constructor and deletes copy constructor (or copy constructor would be allowed and would duplicate fd).
* The destructor would *always* close the fd.

Such a solution requires minimal changes, is very simple (as compared to interesting, but IMHO overly complex `std::variant` version...) yet flexible enough, and it elegantly solves the ownership issues mentioned above:

1. To solve the first issue, we use the fd-move constructor of `UnixFd` when returning `UnixFd` from our D-Bus method. `UnixFd` takes care of closing it.
2. To solve the second issue, `Message& Message::operator>>(UnixFd &item)` would simply invoke the fd-dup constructor so we are independent of the message lifetime.

What do you think?

lechndo:
Hey guys,

unfortunately I have to leave this (and also other discussions) about dbus-cpp because after my evaluation it turned out that our systems are not yet fit for C++17 and the update of the systems will take longer than expected.
The current project is quite urgent so decision was made to stick with our old DBus-C++ lib for now and think again about switching to sdbus-cpp after our systems have been upgraded.

Thx for your support and this great project.
I hope I'll be back in the near future... :wave:

sangelovic:
Ok, thx too for good ideas of yours. But don't postpone the upgrade for too long :) Client tracking feature will be patiently waiting for you ;)

Btw. if you have gcc >= 6, or clang >= 3.7, then this is fully sufficient for C++17 features (notably uncaught exceptions) used by sdbus-c++...

cvuchener:
I liked my variant idea until I had to write the stub generator part. A single always owning UnixFd type would work for me. The downside is you won't be able to send file descriptor you want to keep without an extra dup any more, I guess it is a small price for convenience.

> deletes copy constructor (or copy constructor would be allowed and would duplicate fd).

If you have a copy-from-int constructor and implicit conversion to int, you can copy UnixFd any way. Deleting the copy constructor gives the wrong intention.

Either you make it copyable or move only/always take ownership like unique_ptr, something in-between will be confusing.

What about assignment from int or copy assignment?

UnixFd will also need something like unique_ptr's `release` to move from the UnixFd to whatever the user have to manage their file descriptors.

sangelovic:
I prepped a MR #69 with the above-proposed UnixFd modification. `UnixFd` has a completely new API, and it is a class, not plain struct anymore. It owns the fd, is moveable (transferring the ownership) but also copyable (duplicating the fd). No assignment from int allowed -- `reset()` functionality is provided instead. `release()` is also provided, according to your comments. @cvuchener Would you please review?

PR #69, opened 2019-07-04
Switch from plain UnixFd to owning UnixFd

### LINKED ISSUE #63: When to close a passed file descriptor?

sangelovic:
Fixes #63 
Closes #67

### DIFF
--- a/include/sdbus-c++/Types.h
+++ b/include/sdbus-c++/Types.h
@@ -219,6 +219,7 @@ namespace sdbus {
         {
             close();
             fd_ = ::dup(other.fd_);
+            return *this;
         }
 
         UnixFd(UnixFd&& other)
@@ -231,6 +232,7 @@ namespace sdbus {
             close();
             fd_ = other.fd_;
             other.fd_ = -1;
+            return *this;
         }
 
         ~UnixFd()
--- a/include/sdbus-c++/Types.h
+++ b/include/sdbus-c++/Types.h
@@ -34,6 +34,7 @@
 #include <typeinfo>
 #include <memory>
 #include <tuple>
+#include <unistd.h>
 
 namespace sdbus {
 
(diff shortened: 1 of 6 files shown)


cvuchener:
You are missing "return *this;" in the assignment operators.

------

I tried a test program I wrote for my own patch, and there is an error on the client side with release builds only (works for debug builds of sdbus-cpp):

```
  what():  [System.Error.E2BIG] Failed to deserialize a UnixFd value (Argument list too long)
```

It happens both when I try to receive a fd from the server:

```
#0  0x00007ffff7e258a2 in __cxa_throw () from /lib64/libstdc++.so.6
#1  0x00007ffff7fb2eea in sdbus::Message::operator>>(sdbus::UnixFd&) [clone .cold] () from /home/clement/opt/sdbus-cpp-git/lib64/libsdbus-c++.so.0
#2  0x000000000040c668 in void sdbus::detail::deserialize_pack<sdbus::UnixFd>(sdbus::Message&, sdbus::UnixFd&) ()
#3  0x000000000040bf59 in void sdbus::MethodInvoker::storeResultsTo<sdbus::UnixFd>(sdbus::UnixFd&) ()
#4  0x000000000040b14d in com::example::FileDescriptorTest_proxy::getFile() ()
#5  0x000000000040a6bf in main ()
```

```
    sdbus::UnixFd getFile()
    {
        sdbus::UnixFd result;
        proxy_.callMethod("getFile").onInterface(INTERFACE_NAME).storeResultsTo(result);
        return result;
    }
```

and sending a fd to the server:

```
#0  0x00007ffff7e258a2 in __cxa_throw () from /lib64/libstdc++.so.6
#1  0x00007ffff7fb3f81 in sdbus::MethodCall::sendWithReply() const [clone .cold] () from /home/clement/opt/sdbus-cpp-git/lib64/libsdbus-c++.so.0
#2  0x00007ffff7fb87a4 in sdbus::MethodCall::send() const () from /home/clement/opt/sdbus-cpp-git/lib64/libsdbus-c++.so.0
#3  0x00007ffff7fbcbcd in sdbus::internal::Proxy::callMethod(sdbus::MethodCall const&) () from /home/clement/opt/sdbus-cpp-git/lib64/libsdbus-c++.so.0
#4  0x00007ffff7fb6b1c in sdbus::MethodInvoker::~MethodInvoker() () from /home/clement/opt/sdbus-cpp-git/lib64/libsdbus-c++.so.0
#5  0x000000000040b179 in com::example::FileDescriptorTest_proxy::setFile(sdbus::UnixFd const&) ()
#6  0x000000000040a6a5 in main ()
```

```
    void setFile(const sdbus::UnixFd& file)
    {
        proxy_.callMethod("setFile").onInterface(INTERFACE_NAME).withArguments(file);
    }
```

I don't understand why a UnixFd would be deserialized in the second case.

sangelovic:
Good catch, thanks. I compiled without warnings turned on...

To your issue: the second case is also a UnixFd deserialization failure,
but on the server side. The server tries to read the incoming method call
message, fails and throws sdbus::Error, which is caught and translated into
a D-Bus error reply. The client gets the error reply message and translates
that back to sdbus::Error and re-throws it (because it is the client who
invoked that method and thus should know how to cope with an error). This
happens in MethodCall::sendWithReply, which matches your call stack.

cvuchener:
The bug was actually caused by the missing "return *this", I thought I tested with this fixed, but I must have made a mistake when recompiling/reinstalling. Sorry to have bothered you with that.

Issue #45, opened 2019-03-24
Support for method call timeout


sangelovic:
sdbus-c++ lacks support for user-provided timeouts when calling D-Bus methods. The API should be extended in order for users to be able to provide custom timeouts if needed.

sangelovic:
Fixed in #72

PR #72, opened 2019-08-07
Added support to specify custom timeout value for async DBus client call

lubo-svk:
Currently there is no possibility to specify the timeout value for asynchronous call. This pull request extends the functionality to add support for specifying custom timeout values for asynchronous calls.

### DIFF
--- a/include/sdbus-c++/ConvenienceApiClasses.inl
+++ b/include/sdbus-c++/ConvenienceApiClasses.inl
@@ -485,15 +485,16 @@ namespace sdbus {
     }
 
 
-    inline AsyncMethodInvoker::AsyncMethodInvoker(IProxy& proxy, const std::string& methodName)
+    inline AsyncMethodInvoker::AsyncMethodInvoker(IProxy& proxy, const std::string& methodName, uint64_t timeout)
         : proxy_(proxy)
         , methodName_(methodName)
+        , timeout_(timeout)
     {
     }
 
     inline AsyncMethodInvoker& AsyncMethodInvoker::onInterface(const std::string& interfaceName)
     {
-        method_ = proxy_.createAsyncMethodCall(interfaceName, methodName_);
+        method_ = proxy_.createAsyncMethodCall(interfaceName, methodName_, timeout_);
 
         return *this;
     }
--- a/include/sdbus-c++/ConvenienceApiClasses.inl
+++ b/include/sdbus-c++/ConvenienceApiClasses.inl
@@ -40,6 +40,10 @@
 
 namespace sdbus {
 
+    /*** ----------------- ***/
+    /*** MethodRegistrator ***/
+    /*** ----------------- ***/
+
     // Moved into the library to isolate from C++17 dependency
     /*
     inline MethodRegistrator::MethodRegistrator(IObject& object, const std::string& methodName)
@@ -149,6 +153,9 @@ namespace sdbus {
         return *this;
     }
 
+    /*** ----------------- ***/
+    /*** SignalRegistrator ***/
+    /*** ----------------- ***/
 
     // Moved into the library to isolate from C++17 dependency
     /*
@@ -203,6 +210,9 @@ namespace sdbus {
         return *this;
     }
 
+    /*** ------------------- ***/
+    /*** PropertyRegistrator ***/
+    /*** ------------------- ***/
 
     // Moved into the library to isolate from C++17 dependency
     /*
@@ -309,6 +319,9 @@ namespace sdbus {
         return *this;
     }
 
+    /*** -------------------- ***/
+    /*** InterfaceFlagsSetter ***/
+    /*** -------------------- ***/
 
     // Moved into the library to isolate from C++17 dependency
     /*
@@ -369,6 +382,9 @@ namespace sdbus {
         return *this;
     }
 
+    /*** ------------- ***/
+    /*** SignalEmitter ***/
+    /*** ------------- ***/
 
     // Moved into the library to isolate from C++17 dependency
     /*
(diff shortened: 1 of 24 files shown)


sangelovic:
Hi, thanks for your effort, I'm travelling at the time, I'll look into it
soon.

tools/xml2cpp-codegen/ProxyGenerator.cpp
```diff
@@ -164,16 +172,22 @@ std::tuple<std::string, std::string> ProxyGenerator::processMethods(const Nodes&
         std::string outArgStr, outArgTypeStr;
         std::tie(outArgStr, outArgTypeStr, std::ignore) = argsToNamesAndTypes(outArgs);
 
-        definitionSS << tab << (async ? "void" : retType) << " " << name << "(" << inArgTypeStr << ")" << endl
+        definitionSS << tab << (async ? "virtual void" : retType) << " " << name << "(" << inArgTypeStr << ")" << endl
```
sangelovic: Why is polymorphism (`virtual`) needed here?

include/sdbus-c++/ConvenienceApiClasses.inl
```diff
@@ -485,15 +485,16 @@ namespace sdbus {
     }
 
 
-    inline AsyncMethodInvoker::AsyncMethodInvoker(IProxy& proxy, const std::string& methodName)
+    inline AsyncMethodInvoker::AsyncMethodInvoker(IProxy& proxy, const std::string& methodName, uint64_t timeout)
```
sangelovic: Support for D-Bus call timeout makes a whole lot of sense also for classic sync calls. Would you be so kind to also extend the API of sync calls to support custom timeouts?

include/sdbus-c++/IProxy.h
```diff
@@ -89,7 +90,7 @@ namespace sdbus {
          *
          * @throws sdbus::Error in case of failure
          */
-        virtual AsyncMethodCall createAsyncMethodCall(const std::string& interfaceName, const std::string& methodName) = 0;
+        virtual AsyncMethodCall createAsyncMethodCall(const std::string& interfaceName, const std::string& methodName, uint64_t timeout = 0) = 0;
```
sangelovic: Now that I think of it, sd-bus API is designed such that it doesn't ask for timeout upon message creation, but message dispatch action. Couldn't we model it the same way? So instead of `IProxy::createMethodCall`/`IProxy::createAsyncMethodCall`, we would add an optional timeout parameter in `IProxy::callMethod` virtual methods for both sync and async. (And for those we can of course create a std::chrono::duration non-virtual inline overload, which would cast to usec and delegate to virtual counterpart, in a similar way as proposed above.)

IMO this would make code changes smaller, and wouldn't require keeping timeout as member of `MethodCall`/`AsyncMethodCall`.

src/Proxy.cpp
```diff
@@ -64,9 +64,9 @@ MethodCall Proxy::createMethodCall(const std::string& interfaceName, const std::
...
 }
 
-AsyncMethodCall Proxy::createAsyncMethodCall(const std::string& interfaceName, const std::string& methodName)
+AsyncMethodCall Proxy::createAsyncMethodCall(const std::string& interfaceName, const std::string& methodName, uint64_t timeout)
 {
-    return AsyncMethodCall{Proxy::createMethodCall(interfaceName, methodName)};
+    return AsyncMethodCall{Proxy::createMethodCall(interfaceName, methodName), timeout};
 }
 
 MethodReply Proxy::callMethod(const MethodCall& message)
```
sangelovic: Per suggestion above, we'd have `MethodReply IProxy::callMethod(const MethodCall& message, uint64_t timeout = 0)`, and this timeout would get forwarded to `message.send(timeout)` method, which could be declared like `MethodReply MethodCall::send(uint64_t timeout = 0) const;`

Similarly for async version of these methods/classes (down at Message level, I think we don't need std::chrono::duration overload, uint64_t timeout is sufficient).

tools/xml2cpp-codegen/ProxyGenerator.cpp
```diff
@@ -150,6 +152,12 @@ std::tuple<std::string, std::string> ProxyGenerator::processMethods(const Nodes&
             else if (annotation->get("name") == "org.freedesktop.DBus.Method.Async"
                      && (annotation->get("value") == "client" || annotation->get("value") == "clientserver"))
                 async = true;
+
+           if ((annotation->get("name") == "org.freedesktop.DBus.Method.Async") && (annotation->get("value") == "client")
```
sangelovic: We don't need this `if` at all if we provide timeout support also for sync calls (which we should now that we are adding timeout for async ones). We only need `timeoutValue = annotation->get("timeout");`, and later in the code we just check whether timeoutValue is non-empty string. If it is, we simply add `.withTimeout(timeoutValue)` to the call invocation command.

tools/xml2cpp-codegen/ProxyGenerator.cpp
```diff
@@ -142,6 +142,8 @@ std::tuple<std::string, std::string> ProxyGenerator::processMethods(const Nodes&
 
         bool dontExpectReply{false};
         bool async{false};
+        std::string timeoutValue{"0"};
```
sangelovic: This could simply be `std::string timeoutValue;`, see suggestion below.

sangelovic:
Hi, the thread has been left stale, any plans to continue with this?

lubo-svk:
Hi, currently I am really busy, probably in two weeks I will find some time.

sangelovic:
Fixes #45

PR #76, opened 2019-10-08
Added ability to integrate with foreign event loops

hogliux:
Currently, sdbus-c++ cannot be integrated into foreign event loops, like gtk's or KDE's event loop - and therefore it's impossible to use sdbus-c++ in gtk/kde apps without sdbus-c++ creating an additional thread.

To solve this, sdbus-c++ needs to expose the sdbus' function `sd_bus_get_poll_data`. The result can then be polled for in the user's app or added as an event source to the event loop system the user is using (for example by using the gtk method `g_main_context_add_poll` in gtk apps).

This PR exposes this function and allows the user to process a single sdbus event (whenever the file descriptor signals).

### DIFF
--- a/include/sdbus-c++/IConnection.h
+++ b/include/sdbus-c++/IConnection.h
@@ -46,6 +46,13 @@ namespace sdbus {
     class IConnection
     {
     public:
+        struct PollData
+        {
+            int fd;
+            short int events;
+            uint64_t timeout_usec;
+        };
+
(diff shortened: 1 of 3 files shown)


sangelovic:
Good PR, with nice doxygen docs. The addition makes a lot of sense. Thanks!

hogliux:
Thank you!

PR #77, opened 2019-11-02
Added support for opening a connection to a remote system bus using ssh

jaysterling:
The sd_bus C library provides the ability to open a connection to the system bus on a remote device using ssh. This type of connection is created using the `sd_bus_open_system_remote()` function. This PR adds the necessary plumbing to use this functionality in sdbus-cpp. A new connection creation method was added, `createRemoteSystemBusConnection()` which takes a single argument, the name of the host you'd like to connect to (ex. "user@192.168.1.1").

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -64,6 +64,21 @@ Connection::Connection(Connection::BusType type, std::unique_ptr<ISdBus>&& inter
     loopExitFd_ = createProcessingLoopExitDescriptor();
 }
 
+Connection::Connection(const std::string& host, std::unique_ptr<ISdBus> &&interface)
+    : iface_(std::move(interface))
+    , busType_(BusType::eRemoteSystem)
+    , host_(host)
+{
+    assert(iface_ != nullptr);
+
+    auto bus = openBus(busType_);
+    bus_.reset(bus);
+
+    finishHandshake(bus);
+
+    loopExitFd_ = createProcessingLoopExitDescriptor();
+}
+
 Connection::~Connection()
 {
     leaveProcessingLoop();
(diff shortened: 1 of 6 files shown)


src/Connection.cpp
```diff
@@ -64,6 +64,21 @@ Connection::Connection(Connection::BusType type, std::unique_ptr<ISdBus>&& inter
...
+    , host_(host)
+{
+    assert(iface_ != nullptr);
+
+    auto bus = openBus(busType_);
+    bus_.reset(bus);
+
+    finishHandshake(bus);
+
+    loopExitFd_ = createProcessingLoopExitDescriptor();
```
jaysterling: This code is identical to the other constructor and could be refactored and shared between both, if that's preferred. What's important is that we are able to set `host_` before `openBus()` is called (which makes forwarding this constructor to the other not possible)

sangelovic: I will resolve the duplication in a different way, see general comment below.

sangelovic:
Hi, thanks for contributing!

The change is good in overall. When I started thinking about that constructor code duplication, and about the (un)necessity to keep host name as Connection member, I realized that the current approach to construction of those three types of Connection, that I'd implemented long ago, called for re-thinking. I switched from run-time decision based on bus type enum values to tag dispatching, since we don't need to distinguish among values, but among behaviors of construction. This way, the duplication disappears, the decision which Connection type to construct is resolved at compile time, the code is somewhat simpler and better structured. Alongside, I moved some other stuff around (for example, I applied RAII also to loopExitFd_). Host name doesn't have to be stored as a member, we just need to forward to appropriate sd_bus factory.

I will accept your changes, since they are perfect concerning current design, and I will deliver re-designed Connection construction in a separate commit.

sangelovic:
See 304b69dd8b9c72765ad5ab529783476f2a11bb71

PR #78, opened 2019-11-08
Add a method to retrieve the unique name of a connection

brunomarchand:
Hello!
I added this method to get the unique name of a connection.  It is useful to access remote adapters on unnamed connection.


sangelovic:
Thx!

PR #87, opened 2020-02-20
xml2cpp-codegen: make destructor virtual

DasRoteSkelett:
Even though this is most probably a false warning, you will get
get GCC warning like:
base class ‘class sdbus::ProxyInterfaces<YourProxyHere_proxy>’ has accessible non-virtual destructor [-Wnon-virtual-dtor]
this commit should fix this.

Signed-off-by: Matthias Schoepfer <matthias.schoepfer@ithinx.io>

### DIFF
--- a/tools/xml2cpp-codegen/AdaptorGenerator.cpp
+++ b/tools/xml2cpp-codegen/AdaptorGenerator.cpp
@@ -131,7 +131,7 @@ std::string AdaptorGenerator::processInterface(Node& interface) const
                        << propertyRegistration
          << tab << "}" << endl << endl;
 
-    body << tab << "~" << className << "() = default;" << endl << endl;
+    body << tab << "virtual ~" << className << "() = default;" << endl << endl;
 
     if (!signalMethods.empty())
     {
--- a/tools/xml2cpp-codegen/ProxyGenerator.cpp
+++ b/tools/xml2cpp-codegen/ProxyGenerator.cpp
@@ -96,7 +96,7 @@ std::string ProxyGenerator::processInterface(Node& interface) const
             << registration
             << tab << "}" << endl << endl;
 
-    body << tab << "~" << className << "() = default;" << endl << endl;
+    body << tab << "virtual ~" << className << "() = default;" << endl << endl;
 
     if (!declaration.empty())
         body << declaration << endl;


sangelovic:
Hi :)

Thanks for the suggestion. Generated interface classes have non-virtual destructor by design. The scope of the destructor is `protected`. This together idiomatically means that the destruction should not take place via those interfaces (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c35-a-base-class-destructor-should-be-either-public-and-virtual-or-protected-and-nonvirtual). This is because generated interface classes are not standard polymorphic classes that the client uses to invoke virtual functions. They are only used to provide implementations of virtual functions in derived classes by the client. Hence there is no need to have virtual destructor.

I have never run into the issue you're describing with gcc (currently I'm using gcc 9). I compiled with `-W -Wall -Werror`. What version of gcc are you using? A good compiler should recognize this idiom.

DasRoteSkelett:
Hey! 

Yes, I know that this is not a bug. The compile option is -Wnon-virtual-dtor. We usually compile with -Wall -Wextra -Wshadow -Wnon-virtual-dtor -Wold-style-cast -Woverloaded-virtual -Wuseless-cast -Wstrict-aliasing -Wcast-align -Wconversion -Wsign-conversion -Wunused -Wpedantic -Wdouble-promotion -Wformat=2 -Wduplicated-cond -Wduplicated-branches. GCC version varies around 8.x and 9.2 

It is just to make this warning disappear, so we can see better the real problems that might be there. 
It probably does not hurt do declare the destructor virtual?! 

Regards, 

  Matthias

sangelovic:
Hi, yes it hurts, because the destructor is not meant to be virtual. Making
it virtual is not just an extra keyword, it's an idiomatic change that will
only confuse people. Making destructors in class hierarchies virtual just
"for sure" or to silence down a warning (especially one that we ourselves
control) is a bad thing.

I suggest that you either turn off  -Wnon-virtual-dtor warning globally and
rather use quality static analysis tools to catch uncovered design problems
(my preferred solution, because this warning only confuses people pointing
at destructors that should not by design be virtual) or you turn the
warning off through a pragma directly in code.

sangelovic:
A good static analysis should be smart enough to see that the nonvirtual
destructor is protected, i.e. that the creator intended a
nonvirtual-protected-dtor idiom here. And will not report an issue.

DasRoteSkelett:
Funnily enough, clang-tidy does not complain. Thanks for your time and sdbus-cpp!

PR #89, opened 2020-03-09
Fix #88: Timeout handling.

### LINKED ISSUE #88: Async timeout handling does not work as expected
It seems that call timeouts set for example via `AsyncMethodInvoker::withTimeout()` are not correctly handled by sdbus-cpp's event loop. This can be demonstrated by adjusting test `SdbusTestObject.ThrowsTimeoutErrorWhenMethodTimesOut` in such a way that it also asserts that the expected exception is thrown after roughly 500ms.

As far as I can see, there are two reasons for this:
1. `sd_bus_get_timeout()` actually returns an absolute time point from which one has to subtract the current CLOCK_MONOTONIC time in order to get a relative time that can be passed to `poll()`
1. At the time when the main thread starts an async call C with some small timeout T2, the event loop thread might already be blocked waiting in `poll()` with some timeout T1 > T2. If no other  io-event occurs, this means that the timeout exception for C will be delayed until the call to `poll()` returns.

BenKaufmann:
* Despite what is documented in sd_bus_get_timeout(3), the timeout
  returned is actually an absolute time point of Linux's CLOCK_MONOTONIC
  clock. Hence, we first have to subtract the current time from the
  timeout in order to get a relative time that can be passed to poll.

* For async call timeouts to reliably work, we need a way to notify the
  event loop of a connection that is currently blocked waiting in poll.
  I.e. assume the event loop thread entered poll with a timeout set to
  T1. Afterwards, the main thread starts an async call C with a timeout
  T2 < T1. In order for C to be canceled after its timeout T1 has
  elapsed, we have to be able to notify the event loop so that it can
  update its poll data.

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -40,6 +40,9 @@ namespace sdbus::internal {
 Connection::Connection(std::unique_ptr<ISdBus>&& interface, const BusFactory& busFactory)
     : iface_(std::move(interface))
     , bus_(openBus(busFactory))
+    , loopThreadId_(std::thread::id())
+    , loopExitFd_(EFD_SEMAPHORE)
+    , loopNotifyFd_(0)
 {
     assert(iface_ != nullptr);
 }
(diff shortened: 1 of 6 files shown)


sangelovic:
Hi, thanks for reporting/contributing, will get to this in a few days...

BenKaufmann:
Sorry. This PR is broken as it does not correctly handle multiple connections. I'll open an alternative PR.

PR #90, opened 2020-03-11
Feaure request: Make async call cancellable.

BenKaufmann:
Clients sometimes need to be able to cancel the delivery of the result of an asynchronous call. 

This PR adds support for this use case by extending the existing AsyncCalls logic of Proxy.
In particular, IProxy's async callMethod now returns a PendingCall object representing a handle that can be used to cancel the invocation of a given reply callback.

**NOTE:** The current design does not allow for cancelling from within the user's reply callback as I don't think that this is a sensible thing to do. However, adding support for this would be relatively easy (e.g. by replacing the lock flag with a thread-id and a CAS loop).

Please let me know whether you think this is useful and/or if you see better ways to implement this.

### DIFF
--- a/include/sdbus-c++/ConvenienceApiClasses.h
+++ b/include/sdbus-c++/ConvenienceApiClasses.h
@@ -193,7 +193,7 @@ namespace sdbus {
         template <typename _Rep, typename _Period>
         AsyncMethodInvoker& withTimeout(const std::chrono::duration<_Rep, _Period>& timeout);
         template <typename... _Args> AsyncMethodInvoker& withArguments(_Args&&... args);
-        template <typename _Function> void uponReplyInvoke(_Function&& callback);
+        template <typename _Function> PendingCall uponReplyInvoke(_Function&& callback);
 
     private:
         IProxy& proxy_;
--- a/include/sdbus-c++/ConvenienceApiClasses.inl
+++ b/include/sdbus-c++/ConvenienceApiClasses.inl
@@ -576,7 +576,7 @@ namespace sdbus {
     }
 
     template <typename _Function>
-    void AsyncMethodInvoker::uponReplyInvoke(_Function&& callback)
+    PendingCall AsyncMethodInvoker::uponReplyInvoke(_Function&& callback)
     {
         assert(method_.isValid()); // onInterface() must be placed/called prior to this function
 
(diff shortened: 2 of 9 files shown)


sangelovic:
Hi Benjamin, thanks for your patience, I haven't found time to get to this earlier. Yes, the idea of cancellable async calls is interesting and may surely be useful to the library clients. We should incorporate it. I've quickly scanned your proposed changes, and after a short pondering I think the there is an alternative solution which could actually be simpler. I will dive into this and see whether what I think is true. I'll come back to you asap.

sangelovic:
I've submitted a pull request #95 with an alternative approach to cancellable async calls. Please, feel free to review.

There are some design trade-offs between this (proposed by you here) and the alternative approach:

* Simplicity: The alternative solution needs no additional implementation on top -- no shared/weak pointers (which is good as there is no truly shared ownership semantics), no atomic flags, no cancellation handler implementation.
* Thread safety: The alternative solution is based on TS guarantees already provided by the design of underlying sdbus-c++ layers. For example, releasing an sd-bus slot for an async call needs to lock global sd-bus mutex. But that mutex is also locked when the async callback is invoked, hence the two operations can't happen simultaneously. There is an inherent thread synchronization in those operations. No need for TS guarantees of shared pointers, or atomic flags.
* Run-time cost: Alternative implementation has minimal impact on run-time performance -- especially performance of those invocations that don't truly care about `PendingAsyncCall` value (clients pay almost nothing for what they don't use). Simple two pointers are returned by the `Proxy::callMethod` function, and there are no additional constructions and memory allocations.
* Flexibility: Unlike alternative solution, this approach is more flexible in terms of decoupling the lifetime of `Proxy` instance and `PendingAsynCall` instances created by that `Proxy`. In the alternative approach, clients must assure they don't use lingering `PendingAsyncCall`s after they have destroyed the original `Proxy`. But this is logical in a sense, and actually can help enforce good structural design and lifetime management in client's code. Hence, I consider this flexibility not that useful and not worth sacrificing two design benefits mentioned in above bullet points.

Additionally, `cancel` method in the alternative solution does not return anything. Is it important for you that it returns whether it truly has cancelled the call (especially in face of existence of `isPending` query function)? I'm not sure about that, and, second, the simple modification I could make for this function to return bool would have a slightly weird behavior -- in case the async callback is processed and `cancel` is called simultaneously, `cancel` would return true even when in fact it was too late and the call result was processed, i.e. the call was not truly cancelled. I think your implementation also exhibits that behavior.

Thanks for feedback.

sangelovic:
The discussion continued on #95, and the final merged solution was #96.

PR #91, opened 2020-03-11
Fix #88: Timeout handling.

### LINKED ISSUE #88: Async timeout handling does not work as expected

BenKaufmann:
* Despite what is documented in sd_bus_get_timeout(3), the timeout
  returned is actually an absolute time point of Linux's CLOCK_MONOTONIC
  clock. Hence, we first have to subtract the current time from the
  timeout in order to get a relative time that can be passed to poll.

* For async call timeouts to reliably work, we need a way to notify the
  event loop of a connection that is currently blocked waiting in poll.
  I.e. assume the event loop thread entered poll with a timeout set to
  T1. Afterwards, the main thread starts an async call C with a timeout
  T2 < T1. In order for C to be canceled after its timeout T1 has
  elapsed, we have to be able to notify the event loop so that it can
  update its poll data.


marcullo:
The last commit is my reference (no conflicts with `v1.0.0`). Sorry for messing up.

By the way, it could be nice to have it merged to the master if possible.

riuriuriu:
@sangelovic Did you already look into this? It seems quite important. Otherwise I might have some time this week.

sangelovic:
@riuriuriu No, I didn't have time for deeper look. Would you look into this? It might also be that the fix or a part of it should be also done in sd-bus (e.g. get timeout function which has been reported here to not behave as documented?)

riuriuriu:
Thanks @BenKaufmann for your contribution and apologies for the delay. In general you change seems sound and a very good approach of solving this. Still, I'd like to change a few things like additional error handling, replacing the extra hash map if possible and avoid the one breaking change (return of relative timeout). 
Would it be possible for you to [allow direct changes to your branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)? If not, then I'll just create a new branch and follow-up PR here.

riuriuriu:
> It might also be that the fix or a part of it should be also done in sd-bus (e.g. get timeout function which has been reported here to not behave as documented?)

I think that returning an absolute timeout based of CLOCK_MONOTONIC is the right thing to do. But systemd (and we) will likely have to update the doc. I'll verify it and create a PR on their side.

The reason why the absolute time value works better is, besides to not break the users code, that it allows exactly what this change introduces: Having an additional event "wakeup" source which allows to set a new, earlier timeout. E.g. if another thread performs a new method call with a small timeout value.

BenKaufmann:
@riuriuriu 
> Would it be possible for you to allow direct changes to your branch?

I checked the "Allow edits from maintainers." box. Does it work for you?

anthonybrandon:
I think this can be simplified a bit by adding the eventFd to the connection and passing it to the MethodCall instead of keeping it in the SdBus.  I just copied some code from the current pull request to show the idea below.

<details>
<summary markdown="span">Diff</summary>
```diff
diff --git a/include/sdbus-c++/Message.h b/include/sdbus-c++/Message.h
index 7a48bf9..ce00a3f 100644
--- a/include/sdbus-c++/Message.h
+++ b/include/sdbus-c++/Message.h
@@ -185,6 +185,12 @@ namespace sdbus {
 
         MethodCall() = default;
 
+        MethodCall(void *msg, internal::ISdBus* sdbus, int fd, adopt_message_t)
+               : Message(msg, sdbus, adopt_message)
+               , fd_ { fd }
+       {
+       }
+
         MethodReply send(uint64_t timeout) const;
         void send(void* callback, void* userData, uint64_t timeout, dont_request_slot_t) const;
         [[nodiscard]] Slot send(void* callback, void* userData, uint64_t timeout) const;
@@ -198,6 +204,7 @@ namespace sdbus {
     private:
         MethodReply sendWithReply(uint64_t timeout = 0) const;
         MethodReply sendWithNoReply() const;
+       int fd_ = -1;
     };
 
     class MethodReply : public Message
diff --git a/src/Connection.cpp b/src/Connection.cpp
index b4067d9..7c5ada5 100644
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -96,6 +97,8 @@ void Connection::enterEventLoop()
 
     std::lock_guard guard(loopMutex_);
 
+    eventFd = eventfd(0, EFD_CLOEXEC | EFD_NONBLOCK);
+
     while (true)
     {
         auto processed = processPendingRequest();
@@ -106,6 +109,8 @@ void Connection::enterEventLoop()
         if (!success)
             break; // Exit I/O event loop
     }
+
+    close(eventFd);
 }
 
 void Connection::enterEventLoopAsync()
@@ -221,7 +226,7 @@ MethodCall Connection::createMethodCall( const std::string& destination
 
     SDBUS_THROW_ERROR_IF(r < 0, "Failed to create method call", -r);
 
-    return Message::Factory::create<MethodCall>(sdbusMsg, iface_.get(), adopt_message);
+    return Message::Factory::create<MethodCall>(sdbusMsg, iface_.get(), eventFd, adopt_message);
 }
 
 Signal Connection::createSignal( const std::string& objectPath
@@ -405,8 +410,12 @@ bool Connection::waitForNextRequest()
     assert(loopExitFd_.fd != 0);
 
     auto sdbusPollData = getEventLoopPollData();
-    struct pollfd fds[] = {{sdbusPollData.fd, sdbusPollData.events, 0}, {loopExitFd_.fd, POLLIN, 0}};
+    struct pollfd fds[] = {{sdbusPollData.fd, sdbusPollData.events, 0}, {loopExitFd_.fd, POLLIN, 0}, {eventFd, POLLIN, 0}};
     auto fdsCount = sizeof(fds)/sizeof(fds[0]);
+    if (eventFd == -1) {
+        --fdsCount;
+    }
 
     auto timeout = sdbusPollData.timeout_usec == (uint64_t) -1 ? (uint64_t)-1 : (sdbusPollData.timeout_usec+999)/1000;
     auto r = poll(fds, fdsCount, timeout);
@@ -416,6 +425,12 @@ bool Connection::waitForNextRequest()
 
     SDBUS_THROW_ERROR_IF(r < 0, "Failed to wait on the bus", -errno);
 
+    if (eventFd != -1 && fds[2].revents & POLLIN) {
+        uint64_t value{};
+        auto r = read(eventFd, &value, sizeof(value));
+        SDBUS_THROW_ERROR_IF(r < 0, "Failed to read from the event descriptor", -errno);
+    }
+
     if (fds[1].revents & POLLIN)
     {
         clearExitNotification();
diff --git a/src/Connection.h b/src/Connection.h
index 4a54715..696fa87 100644
--- a/src/Connection.h
+++ b/src/Connection.h
@@ -140,6 +140,7 @@ namespace sdbus::internal {
         std::atomic<std::thread::id> loopThreadId_;
         std::mutex loopMutex_;
         LoopExitEventFd loopExitFd_;
+        int eventFd = -1;
     };
 
 }
diff --git a/src/Message.cpp b/src/Message.cpp
index d59d0af..c24910b 100644
--- a/src/Message.cpp
+++ b/src/Message.cpp
@@ -763,6 +763,10 @@ MethodReply MethodCall::sendWithReply(uint64_t timeout) const
     sd_bus_error sdbusError = SD_BUS_ERROR_NULL;
     SCOPE_EXIT{ sd_bus_error_free(&sdbusError); };
 
+    auto prevTimeout = uint64_t(0);
+    auto bus = ::sd_bus_message_get_bus((sd_bus_message*)msg_);
+    ::sd_bus_get_timeout(bus, &prevTimeout);
+
     sd_bus_message* sdbusReply{};
     auto r = sdbus_->sd_bus_call(nullptr, (sd_bus_message*)msg_, timeout, &sdbusError, &sdbusReply);
 
@@ -771,6 +775,16 @@ MethodReply MethodCall::sendWithReply(uint64_t timeout) const
 
     SDBUS_THROW_ERROR_IF(r < 0, "Failed to call method", -r);
 
+    auto currentTimeout = UINT64_MAX;
+    ::sd_bus_get_timeout(bus, &currentTimeout);
+
+    if (fd_ >= 0 && currentTimeout < prevTimeout)
+    {
+        uint64_t value = 1;
+        const auto cnt = write(fd_, &value, sizeof(value));
+        (void) cnt; // TODO: notifyEventFdLocked: Process write failure
+    }
+
     return Factory::create<MethodReply>(sdbusReply, sdbus_, adopt_message);
 }
 
diff --git a/src/MessageUtils.h b/src/MessageUtils.h
index f8cae77..34d9787 100644
--- a/src/MessageUtils.h
+++ b/src/MessageUtils.h
@@ -57,6 +57,12 @@ namespace sdbus
         {
             return _Msg{msg, sdbus, adopt_message};
         }
+
+        template<typename _Msg>
+        static _Msg create(void *msg, internal::ISdBus* sdbus, int fd, adopt_message_t)
+        {
+            return _Msg{msg, sdbus, fd, adopt_message};
+        }
     };
 
     PlainMessage createPlainMessage();
diff --git a/src/SdBus.cpp b/src/SdBus.cpp
index 75c7bf7..138bb67 100644
--- a/src/SdBus.cpp
+++ b/src/SdBus.cpp
@@ -251,6 +251,27 @@ int SdBus::sd_bus_get_poll_data(sd_bus *bus, PollData* data)
     data->events = static_cast<short int>(r);
 
     r = ::sd_bus_get_timeout(bus, &data->timeout_usec);
+    if (r < 0)
+        return r;
+
+    // Despite what is documented, the timeout returned by ::sd_bus_get_timeout
+    // is an absolute time of linux's CLOCK_MONOTONIC clock.
+    // Values 0 and uint64_t(-1) are used as sentinels meaning
+    // "non-blocking poll" and "no-timeout", respectively.
+    if (data->timeout_usec != UINT64_MAX && data->timeout_usec != 0)
+    {
+        struct timespec ts{};
+        r = clock_gettime(CLOCK_MONOTONIC, &ts);
+        if (r < 0) {
+            return r;
+        }
+
+        const auto now = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::seconds(ts.tv_sec) + std::chrono::nanoseconds(ts.tv_nsec));
+        const auto now_us = static_cast<uint64_t>(now.count());
+        const auto to_us = data->timeout_usec > now_us ? data->timeout_usec - now_us : 0;
+
+        data->timeout_usec = to_us;
+    }
 
     return r;
 }
```
</details>

riuriuriu:
@anthonybrandon thats is my preferred approach too as one of my comments suggest. 

We want an enentfd per connection, and there even is one already there. I'll be able to check that by the end of this week.

riuriuriu:
- Simplified the change by keeping the eventfd in the `Connection` (as @anthonybrandon ) suggested too.
- Don't break user code by keeping the timeout value in the public API (PollData) absolute. 
  - Document PollData
  - Add public methods to convert the absolute time to relative needed by poll.
- updated integrationtests.

Note: Select "Squash and merge" to avoid the rebase conflict.

include/sdbus-c++/Message.h
```diff
@@ -195,9 +196,13 @@ namespace sdbus {
         void dontExpectReply();
         bool doesntExpectReply() const;
 
+    protected:
+        MethodCall(void *msg, internal::ISdBus* sdbus, internal::IConnection* connection, adopt_message_t) noexcept;
+
     private:
         MethodReply sendWithReply(uint64_t timeout = 0) const;
         MethodReply sendWithNoReply() const;
+        internal::IConnection* connection_{};
```
Oxymoron79, line 205: I would assign a `nullptr` here to be more explicit.

src/Connection.cpp
```diff
@@ -221,7 +221,7 @@ MethodCall Connection::createMethodCall( const std::string& destination
 
     SDBUS_THROW_ERROR_IF(r < 0, "Failed to create method call", -r);
 
-    return Message::Factory::create<MethodCall>(sdbusMsg, iface_.get(), adopt_message);
+    return Message::Factory::create<MethodCall>(sdbusMsg, iface_.get(), (sdbus::internal::IConnection*)this, adopt_message);
```
Oxymoron79, line 224: I would prefer a C++-style cast here.

include/sdbus-c++/Message.h
riuriuriu: I kept the style which is already in use, also in the same file. E.g: https://github.com/Kistler-Group/sdbus-cpp/blob/ef6b5ea1384172bf0266a06e1fdea58e07bd3941/include/sdbus-c%2B%2B/Message.h#L172

src/Connection.cpp
riuriuriu: ok, updated

src/Message.cpp
```diff
@@ -771,6 +780,10 @@ MethodReply MethodCall::sendWithReply(uint64_t timeout) const
 
     SDBUS_THROW_ERROR_IF(r < 0, "Failed to call method", -r);
 
+    if (connection_) {
+        connection_->notifyEventLoopNewTimeout();
+    }
```
sangelovic, line 785: @riuriuriu @BenKaufmann Sorry to be late to the party, but do we need to send timeout notification here? It's always a synchronous call...

BenKaufmann:
@sangelovic @riuriuriu Thanks for looking into this. Unfortunately, I don't understand how your solution could work for my use case based on an **external** (asio-based) event loop. In that use case, `IConnection::enterEventLoop{Async}()` is never called and hence `Connection::activeTimeout_` is never set and `Connection::notifyEventLoop()` is never called. Furthermore, even if `Connection::notifyEventLoop()` were to be called, this only notifies the internal (unused in this case) `eventFd_.fd` but not any **external** event loop waiting on the `fd` returned by `IConnection::getEventLoopPollData()`. 

I'm probably missing something, so could you please tell me how one could work with an external loop given only the current interface?

BenKaufmann:
For completeness, my current workaround/hack consists of a new function roughly like this:
```cpp
int Connection::prepareWait()
{
  auto sdbusPollData = getEventLoopPollData();
  auto timeout       = sdbusPollData.getPollTimeout();
  activeTimeout_.store(sdbusPollData.timeout_usec, std::memory_order_relaxed);
  return eventFd_.fd;
}
```

riuriuriu:
We could add the second notification file descriptor to the poll data so you can monitor it too. 
I actually started with that but thought that it's probably too complicated and people won't use it. Would that solve your issue?

BenKaufmann:
I think the fix also misses a crucial `connection_->notifyEventLoopNewTimeout();` in the synchronous call case (`MethodCall::sendWithReply()`). 
Assume we have an (external) event loop waiting in poll for the bus fd to become ready - e.g. maybe waiting for the response of a previous _async_ call to some method `A`. Furthermore, assume at this point a _synchronous_ dbus call to some method `B` is executed. This call could now drain the bus file descriptor reading both the response for `A` as well as for `B`. Only the result for `B` is dispatched as part of the _synchronous_ call, while the response for `A` is still queued. However, without the `connection_->notifyEventLoopNewTimeout();`, the event loop remains blocked in poll because the bus fd is no longer ready for reading.

BenKaufmann:
> We could add the second notification file descriptor to the poll data so you can monitor it too. I actually started with that but thought that it's probably too complicated and people won't use it. Would that solve your issue?

I don't think exposing the second  file descriptor is enough because AFAICS, we also have to make sure that `Connection::activeTimeout_` is correctly set before the external loop calls poll (or similar).

riuriuriu:
Right, that also means that `Connection::activeTimeout_` can't be used. We probably need to go back to the originally proposed approach: read timeout1, send call message, read timeout2, compare timeout1 vs timeout2.

I won't really have the time to follow up on this. @Oxymoron79 or @lubo-svk will you take this issue over? :-)

lubo-svk:
Hello, I can find some time around end of January.

BenKaufmann:
@sangelovic Hi, I'm not sure I can really help here as I don't know all your requirements.
Anyhow, I prepared two simple branches that should solve the issues while staying close to the original code.

- Version 1: https://github.com/Kistler-Group/sdbus-cpp/compare/master...raumfeld:sdbus-cpp:issue-88-1
  - expose additional event fd  
  - expose new function `prepareForEventLoopPoll()`, which users of an external event loop have to call before entering their actual poll function
  - while the implementation is simple (and more or less what we currently use in our integration), having both `prepareForEventLoopPoll()` and `getEventLoopPollData()` in the public API is probably not really obvious

- Version 2: https://github.com/Kistler-Group/sdbus-cpp/compare/master...raumfeld:sdbus-cpp:issue-88-2
  - expose additional event fd  but keep public API (only `getEventLoopPollData()`)
  - let `MethodCall` query and compare the active poll timeout before and after a call
  - the obvious downside of this approach is that it introduces some additional (locking) overhead

AFAICS, a more efficient approach would require extending the `ISdBus` interface and `SdBus` class (e.g. like I did in my initial PR or by adding something like `set_poll_timeout_changed_handler(void(* handler)(sd_bus *bus, void *userdata), void *userdata)`, which `Connection` could then use to notify the event fd.

sangelovic:
@BenKaufmann Hi, thank you for your effort and your proposals. I had time now again to look and think more about it. I would vote for version 2. I think it's for now the best abstraction layer to place the implementation for this into, and does not have the drawback of burdening clients with additional pre-poll methods. What do you think? Would you submit a PR for this?

sangelovic:
@BenKaufmann I thought more about the fundamental cause of necessity of these notifications for poll timeouts. And if we can solve this in a better way, get rid of them.

I realized sd-bus is designed to nicely cope with this timeout issue and our situation it stems from the current design of sdbus-c++ for multi-threading (underlying sd-bus is designed for single-threaded use, in turn). In sdbus-c++, it's possible to send D-Bus messages like async or sync method calls from any thread. That thread sends the message directly, by acessing sd-bus bus instance as a shared resource, bypassing the event loop thread. But that bypassing is the cause here...

If it were the event loop thread that would handle also all outgoing traffic, e.g. would send async D-Bus calls, the poll would be interrupted and the timeout automatically recalculated before the next poll (by calling `getPollData`). It would be more of an asynchronous design -- sdbus-c++ would hand over requests from threads over to the event loop thread, waking it up from poll through a dedicated fd, the event loop thread would carry out the work (e.g., send async D-Bus call, send signal, send async D-Bus method reply...), and go to poll again. The thread would be included and thus timeouts properly recalculated, and not-completely-sent messages sent completely right away (issue #126).

This way, we wouldn't need an very specific solution with event fd just for communicating changing timeouts. The event fd would be there to transfer work items to be carried out by the event loop thread. This would solve async timeout issue you described in this PR.

What do you think? Do you have any thoughts on such an asynchronous model of sdbus-c++?

I am about to refactor sdbus-c++ anyway to some extent. I want to get rid of `eventLoopThreadId_` in `Connection`.

But sync D-Bus method calls are still an issue. With this new asynchronous model (where a request to send the call by writing to event fd would be sent to the event loop thread instead of the thread doing the send directly) we'd get into deadlock in case of a method call done from within the same thread. There it would be simpler to just have synchronous `sd_bus_call` approach.

But that one also causes timeout issue you described in [this comment](https://github.com/Kistler-Group/sdbus-cpp/pull/91#issuecomment-1007211464), we still need those timeout notifications, because of your specific use case with draining poll (if it were multiple threads doing the poll, it would not be an issue because all polls would be waken up. Neither would it be an issue if sync method call would take place from within sdbus-c++ event loop callbacks)... My question here is: Could this be solved on your (client) side? When you are about to go to poll again in your external event loop, you always update the timeout for sdbus-c++ fd by quering sdbus-c++ for it... This solution is done e.g. in sd-bus integration code for sd-event. The prepare callback before the poll always asks for current timeout. But I know in your event loop you may have different sources of wakeups, not just sdbus-c++, and you'd always have to ask for the timeout for sdbus-c++ fd before each poll...

BenKaufmann:
> What do you think? Do you have any thoughts on such an asynchronous model of sdbus-c++?

It sounds nice but I don't understand how it could work with an **external** event loop (e.g. boost asio or GMainLoop) where the Dbus connection is just one of many possible event sources. 

E.g. assume in addition to the Dbus connection with current poll timeout T1, we also have a simple "deadline timer" that expires at some time T2, with T2 < T1. Once the timer expires, we call some dbus method M. It doesn't matter whether M is an asynchronous call or something simple like synchronously getting a remote property. In either case, the poll timeout T1 could now be obsolete (new timeout, drained fd but queued messages) and hence we need a way to "wakeup" any poll on the Dbus Connection's fd so that we can process pending events and/or update the poll timeout. 

AFAICS, Dbus Connection needs a way to communicate with an external event loop. Either by creating and exposing another event fd which the external event loop can also poll or by letting clients register an fd / callback on the connection.

Without such a separate "event" channel, the only alternative I can see right now would be to never expose the actual `sd_bus_get_fd()`/additional event fd but let Dbus Connection always run a poll loop. For the external event loop case, it would then not call `sd_bus_process` though but only notify some file descriptor. The API would remain the same (i.e. `getEventLoopPollData()` returning a single fd) but we'd now have an additional (internal) poll loop.

sangelovic:
@BenKaufmann

> It sounds nice but I don't understand how it could work with an external event loop 

Thanks again Ben for your patience, I have too many things to do in parallel :) But back to your statement -- unless I'm missing something, I think it could work very nicely with external event loops. In the following text I'll elaborate on the two main use cases (and their sub-cases) that we touched in our discussions. In that elaboration I assume the following two things:

1. you have an asio representation of sdbus-c++ connection fd, and an associated asio deadline_timer for timeouts for that sdbus-c++ connection, which takes care of call timeouts.
2. In the handler routine of both the `fd` and the `deadline_timer`, after having iteratively called `IConnection::processPendingRequest()` to process all pending sdbus-c++ events, and before going to `async_wait` again, you get the current timeout value from sdbus-c++ through `IConnection::getEventLoopPollData()` and set that new value in that mentioned `deadline_timer` so you always have up-to-date timeouts reflecting actual state of sdbus-c++ queues (or rather, sd-bus queues underneath) -- this is crucially important because that's how underlying sd-bus was designed to be used, see [here for man page of sd_bus_get_timeout](https://www.freedesktop.org/software/systemd/man/sd_bus_get_timeout.html#) and here for how [sd-bus integrates sd-event by means of prepare callback](https://github.com/systemd/systemd/blob/main/src/libsystemd/sd-bus/sd-bus.c#L3621):

Here the use cases for external event loops:

* Sending an **asynchronous call**
  * **In the context** of the event loop thread (e.g. in a handler routine of an asio deadline timer): No need to do anything here (we just have to get current timeout from sdbus-c++ before the next poll, like mentioned in the assumptions above)
  * In **another** thread (concurrent use of the connection): here, like you said, sdbus-c++ needs a way to wake up the polling event loop thread so we ask the updated timeout from sdbus-c++ and go on polling with it (like mentioned in the assumptions above). But my proposal here was that we use new `eventFd` not for timeout notifications, but to send the async call from that event loop thread.
* Sending a **synchronous call**
  * **In the context** of the event loop thread: Yes, here the poll inside the sync call will drain fd for all incoming D-Bus messages. But when your implementation adheres to the assumptions above, then after the call when you invoke `getEventLoopPollData()` before the next poll, you'll get zero timeout which causes you next poll to wake up immediately, leading you to process the queued messages right away... so this is OK.
  * In **another** thread (concurrent use of the connection): Here we have two threads doing the polling -- the (external) event loop thread (thread A), and the thread doing the synchronous call (thread B). According to my knowledge, both `poll`s should be woken up on an incoming message. But since thread B owns sdbus-c++ mutex, it blocks woken-up thread A from processing the messages. But once thread B gets the reply to his call (or timeout), he relinquishes the mutex and thread A may proceed with processing the queued messages. So this is OK, too.

So to sum up, I can see one problem and one potential problem:

* The problem is like you mentioned in the very original description of this PR -- we have update the timeout on multi-threaded async calls. This we have to solve.
* The potential problem may be that the synchronous calls block the event loop thread from processing other messages that may have arrived in the meantime. But that's the design trait of sd-bus and sdbus-c++ sync call implementation. Users may use it with this potential limitation, but don't need to. They have other choices, too. They may do synchronous call through a ["light-weight", short-lived proxy](https://github.com/Kistler-Group/sdbus-cpp/blob/master/docs/using-sdbus-c%2B%2B.md#using-d-bus-connections-on-the-client-side) that has its own D-Bus connection, thus does not interfere with connection in extenral event loop. Or they may choose to use asynchronous calls for all their calls to avoid blocking behavior.

I hope I was not missing something? Do you have remarks to that? Corrections to my perspective? Thanks.

sangelovic:
@BenKaufmann A little more details on what I meant with using `eventFd` to not send timeout notifications, but directly async calls. My proposal here is -- when clients do the async call, sdbus-c++ internally would store this request in an internal queue, and would do `write` to that `eventFd`. Client's external loop poll would be woken up on that `eventFd`. sdbus-c++'s requirements on clients here would be that on such a wake-up, they call `IConnection::processPendingRequest()` and then go poll again with updated timers with up-to-date timeout from `IConnection::getEventLoopPollData()` (usual stuff). In the process method, sdbus-c++ would also look in that internal "outgoing" queue, and would truly send the async call message at that point. This ensure that that next `IConnection::getEventLoopPollData()` returns correct timeout from the just-before sent async call.

BenKaufmann:
> In that elaboration I assume the following two things:
> 1. you have an asio representation of sdbus-c++ connection fd, and an associated asio deadline_timer for timeouts for that sdbus-c++ connection, which takes care of call timeouts.
> 2. In the handler routine of both the fd and the deadline_timer, after having iteratively called IConnection::processPendingRequest() to process all pending sdbus-c++ events [...]:

This assumption does not address my concern. In my earlier comment, I used the  "deadline timer" only as an example for an **additional** event source, i.e. some event source that is **not at all** concerned with or knows about the dbus connection (therefore never calls `processPendingRequest()/getEventLoopPollData()`) and yet can be dispatched in the same thread (e.g. the main thread). This can easily happen in a heterogeneous system that integrates multiple event sources. Any dbus call invoked in the context of such an additional source would have to somehow "notify" the actual dbus connection loop.

To be honest, I think I simply don't fully understand your proposal and whether (from a client-perspective) it is based on one or two fds. AFAICS, if you really only want to expose exactly one fd, then it boils down to what I have already written:

> the only alternative I can see right now would be to never expose the actual sd_bus_get_fd()/additional event fd but let Dbus Connection always run a poll loop. For the external event loop case, it would then not call sd_bus_process though but only notify some file descriptor.

But this would introduce quite some additional and unnecessary overhead.

Otoh, if you expose two fds (sd_bus_get_fd() + event fd), then everything is possible and from a client-perspective it wouldn't matter whether calls are handled directly in the calling context (notifying event fd on timeout change) or dispatched to the event loop via event fd. An asio-based client would simply run a loop `async_wait`'ing on the two fds as well as a deadline_timer and calling  `processPendingRequest()/getEventLoopPollData()` whenever any of the waits fires.

sangelovic:
@BenKaufmann OK, now I understand your situation better. With classic `poll`, you can always update the d-bus timeouts before the `poll`, regardless of what event (be it d-bus connection itself, d-bus timer, your own deadline timer...) was just processed. But with asio, event sources of more isolated. And doing a D-Bus action on one event may mean modifying the settings (the timeout in this case) of another event source.

The situation you describe (for example, calling an sync/async D-Bus method from some timer event source in the event loop that also incorporates sd-bus connection as an event source, and where connection poll timeout shall then be updated) seems to be more of a problem at sd-bus level. We could create a demo example of yours also with just sd-bus. sd-bus says it supports foreign event loops and even implements integration of sd-event. It provides all required fd's and timeouts, and sdbus-c++ is a just of a thin wrapper here. If a call is made, sd-bus knows new timeouts and could notify its exported "event fd". I'm kind of hesitant to build up workarounds or hacks for this up on sdbus-c++ level.

Have you spoken also to guys who maintain sd-bus? They are available at [systemd mailing list](https://lists.freedesktop.org/archives/systemd-devel/).

There is one case however -- updating timeout on async calls in case of multi-threaded use of the library -- which shall be solved on sdbus-c++ level, because sdbus-c++ supports multi-threading while sd-bus is single-threaded only. I plan to solve this.

PR #94, opened 2020-03-26
Fix #93: Get signals working for multiple proxies.

### LINKED ISSUE #93: Proxy::sdbus_signal_handler() is returning wrong value.
Proxy::sdbus_signal_handler() returns 1 (indicating that no more handlers should be invoked), but I believe it should return 0 instead (causing all other handlers to be invoked).

As the code is currently, listening to a signal works for a single proxy, but when there are multiple proxies listening to the same signal, only one of them will actually be triggered.

```
diff --git a/src/Proxy.cpp b/src/Proxy.cpp
index 4873544..f833e01 100644
--- a/src/Proxy.cpp
+++ b/src/Proxy.cpp
@@ -264,7 +264,7 @@ int Proxy::sdbus_signal_handler(sd_bus_message *sdbusMessage, void *userData, sd

callback(message);

-    return 1;
+    return 0;
}

}
```

olta:
* Proxy::sdbus_signal_handler() needs to return 0 instead of 1 in
  order to allow multiple proxies listening to a signal all being
  triggered.
* Add test for emitting a signal to multiple proxies on same
  connection.

### DIFF
--- a/src/Proxy.cpp
+++ b/src/Proxy.cpp
@@ -238,7 +238,7 @@ int Proxy::sdbus_signal_handler(sd_bus_message *sdbusMessage, void *userData, sd
 
     callback(message);
 
-    return 1;
+    return 0;
 }
 
 }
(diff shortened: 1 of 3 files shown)


sangelovic:
Thank you!

PR #95, opened 2020-03-29
Introduce support for cancellable async calls

sangelovic:


BenKaufmann:
@sangelovic Thank you for taking the time to address this issue.
Regarding your feedback to https://github.com/Kistler-Group/sdbus-cpp/pull/90, to me, it all boils down to your last bullet point there. Namely, do we require lifetime independence of PendingAsyncCall and Proxy or is it ok that the validity of the former implicitly depends on the existence of latter. 
The additional complexity/cost in https://github.com/Kistler-Group/sdbus-cpp/pull/90 is a direct result of the fact that it opts for lifetime independence, which makes the API harder to use incorrectly (in the sense of Scott Meyers' most important design guideline) . However, I agree that this is a design trade-off and that your solution is more efficient.

As a third alternative, we could consider making the lifetime dependency between PendingAsyncCall and IProxy explicit. That is, instead of having `cancel()` and `isPending()` as members of `PendingAsyncCall`, we could make them members of `IProxy`. This way, `PendingAsyncCall` would actually be a opaque cookie that could only be used with the proxy from which it was created.

include/sdbus-c++/IProxy.h
```diff
@@ -263,6 +268,44 @@ namespace sdbus {
...
+        /*!
+         * @brief Answers whether the asynchronous call is still pending
+         *
+         * @return True if the call is pending, false if the call has been fully completed
+         *
+         * Pending call in this context means a call whose results have not arrived,
+         * or have arrived and are currently being handled by the async callback handler.
+         * After the callback handler has completed, the call is deemed completed.
+         */
+        bool isPending();
```
BenKaufmann: I guess this method could be const

src/Proxy.h
```diff
@@ -125,19 +127,38 @@ namespace sdbus::internal {
                 return calls_.emplace(slot, std::move(asyncCallData)).second;
             }
 
-            bool removeCall(void* slot)
+            bool existsCall(void* slot)
```
BenKaufmann: Consider making this const (and the mutex mutable)

sangelovic:
@BenKaufmann thank you for your inputs. I admit I am undecided.

 I can see three candidates at the moment:

* My original approach (A), with implicit dependency between `PendingAsyncCall` and `Proxy`.
* Another alternative (B) that came to my mind is a my approach enriched with lifetime independence ideas of yours, where `Proxy::callMethod` creates `CallData` as a `shared_ptr` (unline `unique_ptr`), and returns `PendingAsyncCall` which would only hold a `weak_ptr`, and `cancel` would get the `shared_ptr`, get access to the proxy instance through `CallData::proxy`, and erase the call item from the map.
* And your interesting alternative (C) with `cancel` and `isPending` functions being part of `Proxy` API.

B is a kind of simplified variant of your original approach, where independence is kept.

I'm contemplating implicit vs no dependency between `PendingAsyncCall` and `Proxy`, in respect to Meyers' guidelines (his books influenced the design of this library API, too, by the way :). With `PendingAsyncCall` having a `weak_ptr`, does this truly lead the client to use the API correctly? It seems more a kind of safety net, being more permissive to the client, and when the user calls `cancel` after `Proxy` has been destroyed, in fact it will do nothing. But is it good to allow clients to destroy `Proxy`s without them caring of potentially leftover handles and cookies to pending calls (among others)? I think the dependency may actually lead to a better, more deterministic code, where (I witnessed and have myself written such a code) the `Proxy` lifetime control is typically localized to the place where it's used and where other related data is stored.

But, most importantly, is implicit dependency such a trouble? Considering that:

* We would provide dependency information in doxygen comments at all the related classes/methods, and in the tutorial. The dependency is also visible in `PendingAsyncCall` class definition.
* The problem only happens when `Proxy` has been destroyed *and* when the client subsequently calls `cancel` or `isPending` on the handle.
* Even if a client wouldn't read docs, and gets to that specific error condition, their CI/CD could/should catch that right away, and the client would fix it properly.

I know we should not rely on client doing this or that... But there is a boundary to everything, and isn't this an acceptable trade-off?

What I think would maybe slightly improve A in terms of expressiveness is if `PendingAsyncCall` were a nested type of `IProxy`.

But the utmost expressive solution (truly alongside Meyers' guidelines IMHO) would be C. Yes, C explicitly demonstrates the lifetime dependency. But... what I don't somehow like about C is that it increases the API size of Proxy, while the object-oriented approach calls for them to be extracted away to the `PendingAsynCall` class. Also, for that to work we must make `Proxy` available in implementations based on generated interfaces (i.e., we must extend `ProxyInterfaces` to provide public `getProxy()` method), which by itself is not a design issue, but the client in this case will have to do something like `myConcatenatorProxy.getProxy().cancel(cookie);` as opposed to `cookie.cancel();`.

So these are the (little longer) lines of my thoughts. Maybe I'm just dived too deep in details. What is your opinion? Your suggestions based on your experience?

BenKaufmann:
@sangelovic 
In your approach A, the handle would be similar to an STL iterator in the sense that the lifetime of iterators is implicitly coupled to their originating container, accessing/dereferencing an iterator is UB if the container is gone, and, given only an iterator, you cannot infer whether it is still valid without additional meta knowledge. Given the ubiquity of iterators in C++, I guess one could therefore say that approach A wouldn't be too surprising for a C++ developer. It is also in line of C++'s mantra of "you don't have to pay for what you don't use". Approach C, makes the dependency more explicit but it feels a little patronizing and also forces every client to maintain an additional reference. Furthermore, once you start talking to different proxies, additional knowledge is required in the client in order to match the handles back to the correct proxy. So, between A and C, I would prefer A.

From my personal experience with async-programming, I still think B is the best (read most flexible) option. There, you often don't have nice and clear hierarchical ownership properties. You often only have a bunch of in flight operations and a large state machine reacting to incoming events.
Approaches similar to B also seem to be quite common in the context of async-programmig, e.g.
glib's GCancellable, folly's CancellationToken, cpprestsdk's cancellation_token etc.

Anyhow, I would be fine with any solution. Currently, I use my implementation of B but for my current project all versions would work.

sangelovic:
@BenKaufmann OK, thanks for inputs, I've decided for B. So it's more flexible for async environments. I've prepared #96 and merged it right away. Let me know in case you'd have additional comments. Thanks a lot for helping to improve the library!

PR #99, opened 2020-04-28
Fixed integration tests for libsystemd older than 242

lejcik:
Recently, a support for method and signal parameter names in introspection was added to the library, but it requires `libsystemd` version 242+ to support it. So that, vtable is set up depending on its version:
https://github.com/Kistler-Group/sdbus-cpp/blob/master/src/VTableUtils.c#L73

However, we are running `libsystemd` 239 and our integration tests are failing, as they expect output with parameter names. Here is a patch that fixes `SdbusTestObject.AnswersXmlApiDescriptionViaIntrospectableInterface` test case to behave correctly depending on `libsystemd` version.


lejcik:
NOTE: see just the last commit, somehow this pull request included more stuff than I wanted 😄

sangelovic:
Hi Vilo, OK, thanks :) But please first rebase your branch on top of this repo's master. Otherwise the diff is a confusing mix, and git won't allow me to merge anyway due to conflicts :)

lejcik:
done, now it should be ready for merge...

PR #102, opened 2020-05-07
Fix #101: sanitize names of namespaces/methods/properties/arguments

### LINKED ISSUE #101: sdbus-c++-xml2cpp generates invalid code
when parameters of methods/signals are named after c++ keywords (e.g. "new") xml2cpp just uses this, which results in invalid c++ code. Example "from the wild":
https://www.freedesktop.org/software/ModemManager/api/latest/gdbus-org.freedesktop.ModemManager1.Modem.html#gdbus-signal-org-freedesktop-ModemManager1-Modem.StateChanged
https://cgit.freedesktop.org/ModemManager/ModemManager/tree/introspection/org.freedesktop.ModemManager1.Modem.xml?id=6c37488d4ebbbb2afdefce193e6fb4070cde8ab3

Proposal: add _ at the end of names that equal reserved keywords in c++ (https://en.cppreference.com/w/cpp/keyword). Eg: new -> new_

I would give this a try myself, just wanted to ask, if this is fine for you or if you have different ideas, how to handle this.

ChristianS99:
…uments

- add a list of c++ keywords and common defines
- sanitize names so that there are no functions/args with names that are reserved in c++


ChristianS99:
Maybe someone can take a look, if I have forgotten some places.

sangelovic:
Thanks, I'll look into this soon.

Issue #104, opened 2020-05-20
Feature request: query objectPath of IProxy/ProxyInterfaces


ChristianS99:
Reasoning:
org.freedesktop.DBus.ObjectManager provides InterfacesAdded/Removed signals, that provide the objectPath. Creating sdbus proxy objects with path is easily doable, but if an object is removed, there is no way to compare the path of the removed object with my sdbus::IProxy objects to remove the corresponding object.

Proposal: implement getPath() (and possibly operator sdbus::ObjectPath()) for (I)Proxy and possibly for ProxyInterfaces.
I can do that, just wanted to ask for comments beforehand.

PS: If I have missed s.th. and this is already possible, please point me to it :)

sangelovic:
One of the ways to solve this would be to keep a dictionary of `Proxy`s, with the key being the object path. But I understand this adds a bit of data duplication, and may not be the acceptable solution in all cases. So I'm OK with providing object path information in the public API, since this information is anyway provided by the client when constructing the instance.

We can also add the `getObjectPath` method to `IObject`, not only `IProxy`. And as you mention correctly, also a method with the same name to `AdaptorInterfaces` and `ProxyInterfaces`, delegating to the underlying object/proxy. I would refrain from conversion operator. And the return value of the metho may be `std::string`, so we are type-consistent with factory methods where it is string. `ObjectPath` as a type is used for values of D-Bus types sent to the bus.

Would you volunteer to prepare a merge request? ;)

ChristianS99:
Yes, will prepare a PR

ChristianS99:
Implemented by ae8849e

PR #105, opened 2020-05-25
Implement #104: add getObjectPath() for classes

ChristianS99:
is @brief enough or should something more added?

### DIFF
--- a/include/sdbus-c++/AdaptorInterfaces.h
+++ b/include/sdbus-c++/AdaptorInterfaces.h
@@ -108,6 +108,14 @@ namespace sdbus {
         }
 
         /*!
+         * @brief return object path of the underlying DBus object
+         */
+        const std::string& getObjectPath() const
+        {
+            return getObject().getObjectPath();
+        }
+
+        /*!
          * @brief Finishes adaptor API registration and publishes the adaptor on the bus
          *
          * This function must be called in the constructor of the final adaptor class that implements AdaptorInterfaces.
--- a/include/sdbus-c++/AdaptorInterfaces.h
+++ b/include/sdbus-c++/AdaptorInterfaces.h
@@ -108,14 +108,6 @@ namespace sdbus {
         }
 
         /*!
-         * @brief return object path of the underlying DBus object
-         */
-        const std::string& getObjectPath() const
-        {
-            return getObject().getObjectPath();
-        }
-
-        /*!
          * @brief Finishes adaptor API registration and publishes the adaptor on the bus
          *
          * This function must be called in the constructor of the final adaptor class that implements AdaptorInterfaces.
@@ -137,6 +129,14 @@ namespace sdbus {
         void unregisterAdaptor()
         {
             getObject().unregister();
+        }
+
+        /*!
+         * @brief Returns object path of the underlying DBus object
+         */
+        const std::string& getObjectPath() const
+        {
+            return getObject().getObjectPath();
         }
 
     protected:
--- a/include/sdbus-c++/IObject.h
+++ b/include/sdbus-c++/IObject.h
@@ -62,11 +62,6 @@ namespace sdbus {
         virtual ~IObject() = default;
 
         /*!
-         * @brief return object path of the underlying DBus object
-         */
-        virtual const std::string& getObjectPath() const = 0;
-
-        /*!
          * @brief Registers method that the object will provide on D-Bus
          *
          * @param[in] interfaceName Name of an interface that the method will belong to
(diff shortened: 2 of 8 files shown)


include/sdbus-c++/AdaptorInterfaces.h
```diff
@@ -107,6 +107,14 @@ namespace sdbus {
         {
         }
 
+        /*!
+         * @brief return object path of the underlying DBus object
+         */
+        const std::string& getObjectPath() const
```
sangelovic, line 113: I'd prefer that we define this function at the end of the class, after `unregisterAdaptor`. This method is not so essential as the other ones for class operation, and these ones I would like to keep up in the class to be visible immediately. Would you please move it there?

include/sdbus-c++/AdaptorInterfaces.h
```diff
@@ -107,6 +107,14 @@ namespace sdbus {
         {
         }
 
+        /*!
+         * @brief return object path of the underlying DBus object
```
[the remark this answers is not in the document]
sangelovic: This applies also to all other comments in this pull request...

include/sdbus-c++/IObject.h
```diff
@@ -61,6 +61,11 @@ namespace sdbus {
     public:
         virtual ~IObject() = default;
 
+        /*!
+         * @brief return object path of the underlying DBus object
+         */
+        virtual const std::string& getObjectPath() const = 0;
```
sangelovic, line 67: Would you please move the declaration of the function to the bottom of the class? This also has the advantage that the binary interface will remain backwards compatible with this change.

include/sdbus-c++/IProxy.h
```diff
@@ -65,6 +65,11 @@ namespace sdbus {
     public:
         virtual ~IProxy() = default;
 
+        /*!
+         * @brief return object path of the underlying DBus object
+         */
+        virtual const std::string& getObjectPath() const = 0;
```
sangelovic, line 71: Ditto, moving to the bottom of the class...

include/sdbus-c++/ProxyInterfaces.h
```diff
@@ -141,6 +141,14 @@ namespace sdbus {
         {
         }
 
+        /*!
+         * @brief return object path of the underlying DBus object
+         */
+        const std::string& getObjectPath() const
+        {
+            return getProxy().getObjectPath();
+        }
```
sangelovic, line 150: Ditto, moving to the bottom of the class...

src/Object.h
```diff
@@ -45,6 +45,11 @@ namespace sdbus::internal {
     public:
         Object(sdbus::internal::IConnection& connection, std::string objectPath);
 
+        const std::string& getObjectPath() const override
+        {
+            return objectPath_;
+        }
```
sangelovic, line 51: The order of methods here doesn't matter much already, but I would also move it after all other virtual methods to keep ordering consistency.

sangelovic, line 51: Additionally, would you please declare the method here and implement it in the `.cpp` file? The method is virtual, so inlining it doesn't truly make sense, the method will be instantiated in several object files and then the linker will anyway have to pick one single place. And anyway, defining it in the `.cpp` file will be consistent with other interface, virtual methods.

src/Proxy.h
```diff
@@ -50,6 +50,11 @@ namespace sdbus::internal {
              , std::string destination
              , std::string objectPath );
 
+        const std::string& getObjectPath() const override
+        {
+            return objectPath_;
+        }
```
sangelovic, line 56: Ditto, defining the method in the `.cpp` file, see comment above...

include/sdbus-c++/AdaptorInterfaces.h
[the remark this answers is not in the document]
ChristianS99: Done

include/sdbus-c++/AdaptorInterfaces.h
ChristianS99: Done

include/sdbus-c++/IObject.h
ChristianS99: Done

include/sdbus-c++/IProxy.h
ChristianS99: Done

src/Object.h
ChristianS99: Done

include/sdbus-c++/ProxyInterfaces.h
ChristianS99: Done

src/Object.h
ChristianS99: Done

src/Proxy.h
ChristianS99: Done

PR #129, opened 2020-10-21
Add ability to set hook for exception propagation

avnik:
Motivation: we need a replace sdbus::Error with custom exception
classes, depending on content of sdbus::Error::getName()

Implementation notes: Hook should return new exception, wrapped
with std::exception_ptr, or nullptr (then original exception re-thrown)

(I tried to make change minimal)


sangelovic:
Hi @avnik, I don't fully understand your motivation, what are you trying to achieve here. From what I see from your changes, I see its quite a special behavior implemented at a specific place, to satisfy your special needs. I think this change does not belong to sdbus-cpp, but onto a higher level - the client. He may decorate his proxies and rethrow whatever exceptions he deems appropriate...

PR #136, opened 2020-12-24
xml2cpp: Add missing EXPAT include dirs

bobrofon:
I was working on a conan.io package for sdbus-c++ and I had a problem to build it with `-DBUILD_CODE_GEN=ON` option, because it couldn't find EXPAT headers.

This patch is required if EXPAT library is installed in non-standard location.
Without `target_include_directories` cmake will find EXPAT library:
  ```
  -- Found EXPAT: .../lib/libexpat.a (found version "2.2.10")
  ```

But `xml.cpp` compilation will fail with error:
  ```
  tools/xml2cpp-codegen/xml.cpp:7:10: fatal error: expat.h: No such file or directory
    7 | #include <expat.h>
  ```


sangelovic:
Thanks for working on conan.io package for sdbus-c++, and for this PR. And sorry for late reaction :)

PR #147, opened 2021-03-03
add getting bus peer credentials (uid/gids)

mbessler:
exposes a subset of the sd_bus_creds_get_* family of
calls to retrieve the bus peer's userid and
group ids.

Example usage:
```c++
void example(sdbus::MethodCall call)
{
    [...]
    SubjectType subject;
    call >> subject;
    std::string bus_name = "";
    auto subject_kind = subject.get<0>();
    if (subject_kind == "system-bus-name") {
        auto subject_details_map = subject.get<1>();
        auto iter = subject_details_map.find("name");
        if (iter != subject_details_map.end()) {
            sdbus::Variant subject_detail_value = iter->second;
            bus_name = subject_detail_value.get<std::string>();
        }
    }
    [...]
    auto uid = call.getCredsUid(bus_name);
    gid_t gid = call.getCredsGid(bus_name);
    std::vector<gid_t> gids = call.getCredsSupplementaryGids(bus_name);
    [...]
}
```

### DIFF
--- a/include/sdbus-c++/IConnection.h
+++ b/include/sdbus-c++/IConnection.h
@@ -31,6 +31,7 @@
 #include <memory>
 #include <chrono>
 #include <cstdint>
+#include <vector>
 
 namespace sdbus {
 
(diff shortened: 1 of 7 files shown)


sangelovic:
Thanks for your effort. There is a pending pull request #151 that will add the requested functionality.

mbessler:
Thank you for getting this cleaned up.

PR #157, opened 2021-03-18
Catch exceptions in the background thread

oleksandrkozlov:
Potentially `enterEventLoop()` may throw an exception. Since the async event loop is running in a background thread, there is no way for clients to catch and handle an exception. So to prevent termination of client's applications all exceptions get caught and silently ignored.

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -107,7 +107,17 @@ void Connection::enterEventLoop()
 void Connection::enterEventLoopAsync()
 {
     if (!asyncLoopThread_.joinable())
-        asyncLoopThread_ = std::thread([this](){ enterEventLoop(); });
+    {
+        asyncLoopThread_ = std::thread([this]()
+        {
+            try
+            {
+                enterEventLoop();
+            } catch (...) {
+                // Ignore unhandled exceptions
+            }
+        });
+    }
 }
 
 void Connection::leaveEventLoop()


oleksandrkozlov:
Hey, Stanislav (@sangelovic), would you mind having a look at these changes?

sangelovic:
Hey Oleksandr.

Looking at your suggested approach with ignoring an exception, I'd say that termination of the application as a result of unhandled exception in a built-in connection event loop thread is a better approach of the two. If you take exception safety guarantees into account... Exception safety at minimum means that we leak no resources and we do not leave the program in an unstable state. `std::terminate` does not leave the program in an unstable state: it does not leave the program running at all.

Whereas, if an exception is masked always, a part of the program stops working somehow without providing any information about the cause (i.e., the exception), which may lead to surprises and an unstable program state.

In general, if sdbus-c++ clients want to have control over handling exceptions, they can simply run the event loop in a thread of their own. There in their thread, they can catch exceptions and react however they want. Or -- they use `enterEventLoopAsync()`, neat and simple call which wraps the underlying built-in thread, but they lose control over exceptions. They have the freedom of choice, and that was the design decision at that time.

But... If we still want to handle exceptions thrown by the built-in connection event loop thread, we have a way to push it a little further. We can store the exception thrown in `enterEventLoopAsync()` and then re-throw it once the built-in thread joins. Would you agree with such an approach? If yes, would you volunteer to implement that in this PR?

sangelovic:
I don't much like bending (obscuring) the API on several layers for some special exceptional cases. Rather then adding callback parameters on several layers for special treatments, I'd favor keeping the potential exception and handing it over upon joining. We can do that by marking the connection destructor (and Proxy's, too) `noexcept(false)`. The client would thus know that those destructors may throw (the docs would explain in which cases). And/Alternatively, provide `Proxy::getConnection()` method. With this, and already existing `Connection::leaveEventLoop()` method, the client would have full control over whether they don't care about throwing destructors, or they take it into their hands and exit connection's event loop by their own explicit call -- which can throw, but clients can expect that already (and catch the exception and process it).

oleksandrkozlov:
My concern is that in a good design "a destructor must not fail" [1]. But I got your point. Anyway, I can solve my issue simply by not using `enterEventLoopAsync()` as you suggested. I ended up with the below wrapper (simplified version). That's what feel free to close this pull request and thank you for your time.

```cpp
struct client {
  client(std::string destination, std::string object_path)
    : proxy{sdbus::createProxy(*connection, std::move(destination)
    , std::move(object_path))}
    , loop_thread{[this] {
        try {
          connection->enterEventLoop();
        } catch (...) {}
    }} {}

  ~client() noexcept {
    try {
      connection->leaveEventLoop();
      loop_thread.join();
    } catch (...) {}
  }

  std::unique_ptr<sdbus::IConnection> connection{
      sdbus::createSystemBusConnection()};
  std::unique_ptr<sdbus::IProxy> proxy;
  std::thread loop_thread;
};
```

* [1] https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c36-a-destructor-must-not-fail

sangelovic:
Yes, I understand throwing d-tors are dangerous, in double-exception situations... So we don't necessarily need to mark them as throwing. We could simply provide mechanisms for the other option for the client to handle that, and leave the case that throwing upon joining in a destructor would terminate the program. I've created an issue to consider that.

PR #167, opened 2021-03-24
Add API to get message path

m-blaha:


include/sdbus-c++/Message.h
```diff
@@ -131,6 +131,7 @@ namespace sdbus {
         std::string getInterfaceName() const;
         std::string getMemberName() const;
         std::string getSender() const;
+        std::string getPath() const;
```
dleeds-cpi, line 134: While you're here, it might be nice to also add:
```c++
std::string getDestination() const;
```

sangelovic: @m-blaha Would you be that kind to also add `getDestionation()`? Or shall we rather create a separate PR for that?

m-blaha: Of course.
Also I added handling the situation when the message does not have some fields (interface, member, path and destination) set. In this situation the getters from underlying sdbus library return NULL.

m-blaha: Or would it be better to just return empty string in such situations? I do not like getters throwing an exceptions.

dleeds-cpi (approved):

include/sdbus-c++/Message.h
sangelovic: @m-blaha I agree. If it were a true erroneous condition that a sd-bus message wouldn't have any of those fields, then I guess sd-bus API would return an error code. But it just returns `const char*`, where `nullptr` means no data. It seems to be a kind of possible, valid value. So -- I agree -- in our library, we could represent that sentinel value by an empty string, as @dleeds-cpi suggested, rather than throwing exceptions which in sdbus-c++ are used for rather serious, unexpected situations.

src/Message.cpp
```diff
@@ -583,19 +583,37 @@ void Message::rewind(bool complete)
 
 std::string Message::getInterfaceName() const
 {
-    return sd_bus_message_get_interface((sd_bus_message*)msg_);
+    auto interface = sd_bus_message_get_interface((sd_bus_message*)msg_);
```
sangelovic, line 586: If we go for empty strings for nonexisting sd-bus message data, we could have this here (and in other methods in this PR):

```suggestion
    auto interface = sd_bus_message_get_interface((sd_bus_message*)msg_);
    return interface != nullptr ? {interface} : {};
```

src/Message.cpp
```diff
@@ -583,19 +583,37 @@ void Message::rewind(bool complete)
...
+    SDBUS_THROW_ERROR_IF(!interface, "Failed to get message interface", EINVAL);
+    return interface;
```
sangelovic, line 588: If we go for empty strings for nonexisting sd-bus message data, we could have this here (and in other methods in this PR):

```suggestion
    auto interface = sd_bus_message_get_interface((sd_bus_message*)msg_);
    return interface != nullptr ? {interface} : {};
```

m-blaha: Nice, but unfortunately does not work for me (gcc version 10.3.1):
```
sdbus-cpp/src/Message.cpp: In member function ‘std::string sdbus::Message::getInterfaceName() const’:
sdbus-cpp/src/Message.cpp:587:35: error: expected primary-expression before ‘{’ token 
  587 |     return interface != nullptr ? {interface} : {};
      |                                   ^
sdbus-cpp/src/Message.cpp:587:34: error: expected ‘:’ before ‘{’ token
  587 |     return interface != nullptr ? {interface} : {};
      |                                  ^~
      |                                  :
sdbus-cpp/src/Message.cpp:587:35: error: expected primary-expression before ‘{’ token 
  587 |     return interface != nullptr ? {interface} : {};
```

sangelovic: OK, I see, seems not to work with ternary operator. Your approach is fine. Thanks

PR #168, opened 2021-03-24
Add API to set signal destination

m-blaha:
With this patch it is possible to create unicast signals - e.g. I want to keep client informed about the progress in processing of his request using regularly sent dbus signals so that the client could print a progress bar or something. Only the client that sent original request is interested in such signal:

```
auto signal = dbus_object->createSignal(interface, "ProcessProgress");
signal.setDestination(request_sender);
signal << items_processed;
signal << items_total;
dbus_object->emitSignal(signal);
```

### DIFF
--- a/include/sdbus-c++/Message.h
+++ b/include/sdbus-c++/Message.h
@@ -215,6 +215,7 @@ namespace sdbus {
 
     public:
         Signal() = default;
+        void setDestination(const std::string& destination);
         void send() const;
     };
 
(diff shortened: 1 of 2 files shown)


dleeds-cpi:
This is a nice feature to have. As it turned out, I also suggested it just after you put up this PR. https://github.com/Kistler-Group/sdbus-cpp/discussions/169

sangelovic:
Thanks :)

PR #170, opened 2021-03-30
fix #145, signals are not filtered by sender

### LINKED ISSUE #145: Signals are not filtered by sender
The `class ProxyInterfaces`'s constructor or the `createProxy(..)` take a destination as argument. This destination is never used as filter for received signals. This results in signals being received from **all** services on that bus. 

The filter is setup here:
https://github.com/Kistler-Group/sdbus-cpp/blob/master/src/Connection.cpp#L420
and should use e.g. `sender='org.example.App2'`.

I'll create a Pull Request as soon as I have time to fix this properly.

riuriuriu:
fix for #145 

A new integration test is added first with uncovers the issue. It shows that a proxy gets the signal from any/all bus names (destinations) if the object path and interface match. We want the proxy to only connect to the `destination` which was given as argument during construction.

A missing `sender=..` in the match filter rule solves the issue.

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -289,7 +289,8 @@ void Connection::emitInterfacesRemovedSignal( const std::string& objectPath
     SDBUS_THROW_ERROR_IF(r < 0, "Failed to emit InterfacesRemoved signal", -r);
 }
 
-SlotPtr Connection::registerSignalHandler( const std::string& objectPath
+SlotPtr Connection::registerSignalHandler( const std::string& sender
+                                         , const std::string& objectPath
                                          , const std::string& interfaceName
                                          , const std::string& signalName
                                          , sd_bus_message_handler_t callback
@@ -297,7 +298,9 @@ SlotPtr Connection::registerSignalHandler( const std::string& objectPath
 {
     sd_bus_slot *slot{};
 
-    auto filter = composeSignalMatchFilter(objectPath, interfaceName, signalName);
+    // alternatively to our own composeSignalMatchFilter() implementation, we could also use sd_bus_match_signal() from
+    // https://www.freedesktop.org/software/systemd/man/sd_bus_add_match.html#
+    auto filter = composeSignalMatchFilter(sender, objectPath, interfaceName, signalName);
     auto r = iface_->sd_bus_add_match(bus_.get(), &slot, filter.c_str(), callback, userData);
 
     SDBUS_THROW_ERROR_IF(r < 0, "Failed to register signal handler", -r);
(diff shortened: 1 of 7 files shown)


src/Connection.cpp
```diff
@@ -289,15 +289,18 @@ void Connection::emitInterfacesRemovedSignal( const std::string& objectPath
...
                                          , const std::string& interfaceName
                                          , const std::string& signalName
                                          , sd_bus_message_handler_t callback
                                          , void* userData )
 {
     sd_bus_slot *slot{};
 
-    auto filter = composeSignalMatchFilter(objectPath, interfaceName, signalName);
+    // alternatively to our own composeSignalMatchFilter() implementation, we could also use sd_bus_match_signal() from
+    // https://www.freedesktop.org/software/systemd/man/sd_bus_add_match.html#
```
phatina, line 302: Don't we want do it now?

phatina (approved):
Works for me ;)

src/Connection.cpp
riuriuriu: It was just the smaller change but it's probably better to use the convenience function if it is already provided by the underlying sdbus lib. Especially if it would have avoided this bug. 

We state that the minimum libsystemd version is v236. Let me check if this function is there in v236. If yes, I will update the code.

riuriuriu: v237 introduced sd_bus_match_signal(): https://github.com/systemd/systemd/commit/b423e4fb73866e529869b348efb7169ee91f00c9

I don't think this justifies to increase the libsystemd dependency from v236 to v237. The comment was updated but under this circumstances I'd prefer to keep as is.

phatina: Okay, agreed then.

dleeds-cpi:
@riuriuriu I have in the past observed issues where the `sender` field is populated with the unique bus ID, and not the service's well-known name. If this were to occur, these changes would filter out all signals. Unfortunately, I do not recall exactly what version of the D-Bus Daemon was in use, or whether the offending signals were emitted by `sd-bus` or a different D-Bus library.

Have you tested these changes with a variety of client library versions and D-Bus daemons?

riuriuriu:
@dleeds-cpi The concrete issue reported by #145 was fixed here. Bug reports are very welcome but you should open a new issue and describe exactly under which circumstances the the problem occurs and be sure that it's caused by sdbus-cpp.

sangelovic:
@dleeds-cpi As an additional comment, sd-bus since some recent version also provides a convenient API function `sd_bus_match_signal` which takes `sender` as its parameter. It makes logical sense. For some specific cases like ones you mentioned, the feature you proposed [here](https://github.com/Kistler-Group/sdbus-cpp/discussions/258#discussioncomment-2807539) should give clients the tool to solve the thing themselves.

PR #172, opened 2021-04-07
Provide CMake config and PkgConfig files for tools

granxarixia:
When a project uses automatically generated interface headers using the stub generator it is convenient if the conversion by the stub generator can be integrated into the project itself. Until now it was not possible to search for the availibility of the stub generator and it had to be invoked blindly.

This commit provides CMake and PkgConfig files which allow projects to check availibility of the stub generator as well as its version and to automatically resolve its installation path.


sangelovic:
Thanks Michal :)

PR #180, opened 2021-05-06
Enable default construction of PendingAsyncCall

dleeds-cpi:
This is helpful in use cases where a user defined class wants to store a `PendingAsyncCall` as a member variable, or in a STL container.

I mentioned this change here: https://github.com/Kistler-Group/sdbus-cpp/discussions/169

### DIFF
--- a/tests/integrationtests/DBusAsyncMethodsTests.cpp
+++ b/tests/integrationtests/DBusAsyncMethodsTests.cpp
@@ -201,6 +201,22 @@ TEST_F(SdbusTestObject, AnswersThatAsyncCallIsNotPendingAfterItHasBeenCompleted)
     ASSERT_TRUE(waitUntil([&call](){ return !call.isPending(); }));
 }
 
+TEST_F(SdbusTestObject, AnswersThatDefaultConstructedAsyncCallIsNotPending)
+{
+    sdbus::PendingAsyncCall call;
+
+    ASSERT_FALSE(call.isPending());
+}
+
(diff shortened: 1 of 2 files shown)


tests/integrationtests/DBusAsyncMethodsTests.cpp
```diff
@@ -201,6 +201,23 @@ TEST_F(SdbusTestObject, AnswersThatAsyncCallIsNotPendingAfterItHasBeenCompleted)
...
+    // Assign
+    call = m_proxy->doOperationClientSideAsync(100);
+
+    ASSERT_TRUE(call.isPending());
+
+    call.cancel();
+
+    ASSERT_FALSE(call.isPending());
+}
+
```
sangelovic, line 220: I would suggest to split this into two test cases, each having a single responsibility, with names revealing behaviors under tests, and with more of "arrange-act-assert" ("given-when-then") structure:

```suggestion
TEST_F(SdbusTestObject, AnswersThatDefaultConstructedAsyncCallIsNotPending)
{
    sdbus::PendingAsyncCall call;
    
    ASSERT_FALSE(call.isPending());
}

TEST_F(SdbusTestObject, SupportsAsyncCallCopyAssignment)
{
    sdbus::PendingAsyncCall call;

    call = m_proxy->doOperationClientSideAsync(100);

    ASSERT_TRUE(call.isPending());
}

```

dleeds-cpi: Thanks, will do.

PR #182, opened 2021-05-09
Add createDefaultBusConnection()

dleeds-cpi:
This internally calls `sd_bus_open()`, which automatically selects the system or session bus connection based on the presence and content of a `DBUS_STARTER_BUS_TYPE` environment variable and whether the calling process has root privileges.

This option is very helpful when creating services and clients that will use the system bus in production, but connect to a session for testing.

Note: it was desirable to change `createConnection()` to open the default bus instead of creating a new function, but I assumed this would be too controversial, as it could break existing code that relies on it always opening the system bus.

**Additional changes**
* Removed assertions null-checking `make_unique()` return values. `make_unique()` calls `new`, and `new` is expected to throw or abort on failure, making the assertions unhelpful.
* Corrected a typo in the `ClosesAndUnrefsBusWhenDestructed` unit test for the system bus (tested the wrong bus).

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -44,6 +44,11 @@ Connection::Connection(std::unique_ptr<ISdBus>&& interface, const BusFactory& bu
     assert(iface_ != nullptr);
 }
 
+Connection::Connection(std::unique_ptr<ISdBus>&& interface, default_bus_t)
+    : Connection(std::move(interface), [this](sd_bus** bus){ return iface_->sd_bus_open(bus); })
+{
+}
+
 Connection::Connection(std::unique_ptr<ISdBus>&& interface, system_bus_t)
     : Connection(std::move(interface), [this](sd_bus** bus){ return iface_->sd_bus_open_system(bus); })
 {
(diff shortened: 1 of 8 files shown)


dleeds-cpi:
@sangelovic / @riuriuriu would you like to review or close this PR? Thanks!

src/Connection.cpp
```diff
@@ -484,10 +489,23 @@ std::unique_ptr<sdbus::IConnection> createConnection(const std::string& name)
...
+std::unique_ptr<sdbus::IConnection> createDefaultBusConnection()
+{
+    auto interface = std::make_unique<sdbus::internal::SdBus>();
+    constexpr sdbus::internal::Connection::default_bus_t default_bus;
+    return std::make_unique<sdbus::internal::Connection>(std::move(interface), default_bus);
+}
+
+std::unique_ptr<sdbus::IConnection> createDefaultBusConnection(const std::string& name)
+{
+    auto conn = createConnection();
```
sangelovic, line 501: Shouldn't we call `createDefaultBusConnection()` here?

sangelovic (changes requested):
Hi @dleeds-cpi , I got to your PR now, thanks for pinging me :) But I shall have some more time in the coming months, so please feel free to proceed with the contribution of your ideas/improvements... :)

src/Connection.cpp
dleeds-cpi: Yes, good catch. This was left over from when I had `createConnection()` creating the default bus connection.

sangelovic (approved):

PR #190, opened 2021-06-08
Provide access to D-Bus message in high-level API

### LINKED ISSUE #71: How to access to sender address
How to access sender address or get another message header info in adapter?

### LINKED ISSUE #146: Find sender of method invoke
I need to find out who is the sender of a certain dbus method invoke.

So I have a service and a client created via sdbus. I am using the XML approach to create all the glue code inside c++.
How can I find out from a registered method who has invoked it?

I found this
https://github.com/Kistler-Group/sdbus-cpp/blob/master/include/sdbus-c%2B%2B/Message.h#L132

But I have no idea how to access this Message class from my automatically generated Adapter class.

I found also this issue
https://github.com/Kistler-Group/sdbus-cpp/issues/71
But no Solution atm.

sangelovic:
This PR implements access to the corresponding D-Bus message in client's callback handlers written on high-level (convenience) API.

Fixes #146 
Fixes #71


sangelovic:
@riuriuriu Would you please review?

sangelovic:
@riuriuriu ping :)

riuriuriu:
@sangelovic  I'm on parental leave and will have more time from next Monday on :-).

In general I agree with your reasoning in #71, specifically that the convenience and stubs API should abstract away the messages. If someone still requires them nevertheless, it might be a warning sign for an anti-pattern.

On the other hand, I understand that the message can provide additional helpful information (at least for debugging, maybe more). So I'm OK with providing access if it's done in a safe and intuitive way.

That said, I'm a bit concerned about this:
> Both IObject and IProxy provide the getCurrentlyProcessedMessage() method. This method is meant to be called from within a callback handler. It returns a pointer to the corresponding D-Bus message that caused invocation of the handler. The pointer is only valid (dereferencable) as long as the flow of execution does not leave the callback handler. When called from other contexts/threads, the pointer may be both zero or non-zero, and its dereferencing is undefined behavior.

I would prefer that the API already prevents a wrong usage. What about just having an `IMessageInfo` as optional last argument to any callback? I'm not sure yet how that would blow the TypeTraits or if it's even possible. So the "normal" user would implement the concatenator example callback as before:
```cpp
auto concatenate = [&concatenator](const std::vector<int> numbers, const std::string& separator) { /*....*/}
```
And someone interested in additional message info could add a `const IMessageInfo&` to the callback arguments:
```cpp
auto concatenate = [&concatenator](const std::vector<int> numbers, const std::string& separator, 
const IMessageInfo& message) { /*....*/}
```
`IMessageInfo` would provide just an informational subset like the `getSender()` method. It is only made available when the callback is called and the reference should make it clear that it shouldn't be used after the callback returns. Alternatively to the reference, we could use a std::shared_ptr.

Does that make any sense?

sangelovic:
@riuriuriu Yes, what you're proposing makes sense to me, and I considered that option at that time, too. However, AFAIK, it's not technically possible. We can't specialize variadic templates for type traits based on the last argument type. Even if we were, there is more to it -- we'd have to somehow extend XML syntax to allow generation of implementations of methods with that extra parameter. This would blow the complexity of this solution just to achieve a quite simple thing.

That simple thing is there during processing, in `Proxy`, or `Object` respectively, we just need a way to access it.

In view of this, I understand the limitations of `getCurrentlyProcessedMessage()`, but I see it as a good trade-off. First, introducing this method does in no way affect current client implementations. Second, whoever needs this special method (a minority of clients I guess), will look anyway at the docs searching for it, where they'll read the context and (kind of logical) limitations of the method. I don't deem this as *unsafe* per se, since the method's behavior is always defined (although its result unspecified in unsupported case) and it would never crash, even if used incorrectly in multi-threaded environment. If the clients use this method and its results improperly and carelessly (e.g. they dereference a nullptr), they don't understand what they are doing, and then I guess they have bigger problems, and while I think having API leading to expressive, robust, safe code for 9x% of client's use cases is worthwhile, safeguarding every stupid usage in every single special corner of the library isn't.

Issue #135, opened 2020-12-14
Signal::peekType() throws segfault


GeodeLX:
I have something like the following code which I'm using with sdbus-c++ v0.8.3:
`
void onInterfacesAdded(sdbus::Signal& signal)
{
    std::cerr << "Interface added" << std::endl;
    std::string strType("<type?>");
    std::string strContents("<contents?>");
    signal.peekType(strType, strContents);
    std::cerr << "  PeekType(" << strType << ", " << strContents << ")" << std::endl;
}

void anotherFunction(const char* destinationName, const char* obectPath) {
    auto myProxy = sdbus::createProxy(destinationName, objectPath);
    myProxy->registerSignalHandler("org.freedesktop.DBus.ObjectManager", "InterfacesAdded", &onInterfacesAdded);
    myProxy->finishRegistration();
    while(1);
}
`
When I get to signal.peekType(), the code throws a segmentation fault.  A bit of digging indicates that sd_bus_message_peek_type appears to be setting contentsSig to nullptr, and this causes a seg fault when the contentsSig is copied into contents (a std::string):
`// extract from Message.cpp
// Message::peekType(...)
    auto r = sd_bus_message_peek_type((sd_bus_message*)msg_, &typeSig, &contentsSig);
    contents = contentsSig;
`

PR #191, opened 2021-06-12
Fix issue #135: Segfault in Message::peekType()

BenKaufmann:
* Add missing nullptr check in Message::peekType().

* According to its specification, sd_bus_message_peek_type() sets a
  given contents parameter to NULL if the next element in a message is
  not a container. Since assigning a nullptr to a std::string has
  undefined behaviour (typically resulting in an invalid memory access),
  Message::peekType() must not assign contentsSig unconditionally.


sangelovic:
Thanks @BenKaufmann !

PR #192, opened 2021-06-14
docs: Add information about conan recipe

bobrofon:
I'm not sure but it may be helpful to have reference to Conan package of sdbus-c++ in documentation.


sangelovic:
Very well! Thanks :)

Issue #178, opened 2021-05-03
Consider transferring exceptions from internal event loops


sangelovic:
The proposal is laid out in the discussion in this merge request: https://github.com/Kistler-Group/sdbus-cpp/pull/157

We can consider to use `std::exception_ptr` to transfer exceptions from the sdbus-c++ internal event loop thread out to the main thread upon joining. We can keep the destructors noexcept, so exceptions in this case would terminate the program, but provide the other option, too: clients can design their code so that master thread joins with the event loop thread explicitly before invoking a destructor of proxy/connection. For that, we need to extend `IProxy` to provide `getConnection()` method. Then clients have the freedom to react to potential exceptions.

dleeds-cpi:
There is a subset of conditions that may throw exceptions where catching the exception within the library is appropriate.

Specifically, if the exception is thrown due to an asynchronously received incoming message not having the expected payload, it should be ignored. This is a use case where no action can be taken by the recipient of a malformed message to mitigate the issue, and it should just be dropped.

This should be done in the following places:
1. `SignalSubscriber::call`
2. `AsyncMethodInvoker::uponReplyInvoke()`

sangelovic:
It's actually not a good idea that sdbus-c++ exceptions propagate all the way to the event loop itself. This would mean propagating through libsystemd, which is a C library, and this is utterly unsafe. Hence, in the proposed #196, these exceptions are caught in the `Proxy` handlers. Other exceptions are ignored (the clients should take care they are not thrown, or their emission means something has gone very much wrong).

PR #193, opened 2021-06-17
Avoid propagating msg unpack exceptions to event loop

dleeds-cpi:
This change addresses conditions where an exception is thrown by the library upon receipt of a malformed message from an external source, and propagated up to the event loop with no chance of interception by the application. This issue is only experienced by proxy convenience APIs, as low-level APIs allow the application to unpack the message.

Strategy:
1. For malformed signals received by proxies: ignore the signal.
2. For malformed async method responses, translate the unpack exception into an sdbus::Error, and pass it to the caller as expected.

### DIFF
--- a/include/sdbus-c++/ConvenienceApiClasses.inl
+++ b/include/sdbus-c++/ConvenienceApiClasses.inl
@@ -588,7 +588,20 @@ namespace sdbus {
 
             // Deserialize input arguments from the message into the tuple (if no error occurred).
             if (error == nullptr)
-                reply >> args;
+            {
+                try
+                {
+                    reply >> args;
+                }
+                catch (const sdbus::Error& e)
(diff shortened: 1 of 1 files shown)


sangelovic:
Fixes #178

sangelovic:
Thank you @dleeds-cpi for your contribution

sangelovic:
@dleeds-cpi I took some time to think about handling of these errors, as well as event loop errors in general, in sdbus-c++. I have come to the conclusion that your proposal makes most sense. It may generally be unsafe to throw exceptions from C++ handlers invoked by underlying C library libsystemd -- I'm not sure if the unwinding mechanism would leave libsystemd in a valid state. We should probably not allow sdbus-c++ to throw exceptions there.

PR #197, opened 2021-07-22
Support Error parameter in signal handlers

sangelovic:
Currently, signal handlers are not invoked (i.e., the signal is silently dropped) if there is a signal signature mismatch. This PR enables clients to be informed of such situations (e.g. they make a mistake in signal's signature). So with this, if clients want to be informed of such situations, they can prepend `const sdbus::Error*` parameter to their signal handler's parameter list. This argument will be `nullptr` in normal cases, and will provide access to the corresponding `sdbus::Error` exception in case of deserialization failures

### DIFF
--- a/include/sdbus-c++/ConvenienceApiClasses.inl
+++ b/include/sdbus-c++/ConvenienceApiClasses.inl
@@ -640,11 +640,34 @@ namespace sdbus {
             // as a storage for the argument values deserialized from the signal message.
             tuple_of_function_input_arg_types_t<_Function> signalArgs;
 
-            // Deserialize input arguments from the signal message into the tuple
-            signal >> signalArgs;
+            // The signal handler can take pure signal parameters only, or an additional `const Error*` as its first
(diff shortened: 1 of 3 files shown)


include/sdbus-c++/ConvenienceApiClasses.inl
```diff
@@ -640,21 +640,34 @@ namespace sdbus {
...
-            catch (const sdbus::Error& e)
+            else
             {
-                // The convenience API callback cannot handle an incoming signal with
-                // an unexpected payload, so catch and ignore this exception to avoid
-                // propagating it up the call stack to the event loop.
-                return;
-            }
+                // Deserialize input arguments from the signal message into the tuple
+                signal >> signalArgs;
```
dleeds-cpi, line 666: Looks like we need another `try`/`catch` here to match the behavior in the above comment.

dleeds-cpi: Ah, never mind. I see you've addressed this in https://github.com/Kistler-Group/sdbus-cpp/pull/196

PR #198, opened 2021-07-22
Fix potential race condition in Object destruction

### LINKED ISSUE #189: Data race in Object class
The same as #133 , except that it applies to `Object` class

sangelovic:
This mirrors https://github.com/Kistler-Group/sdbus-cpp/pull/177.

Fixes #189

### DIFF
--- a/src/Object.cpp
+++ b/src/Object.cpp
@@ -73,7 +73,7 @@ void Object::registerMethod( const std::string& interfaceName
 {
     SDBUS_THROW_ERROR_IF(!methodCallback, "Invalid method callback provided", EINVAL);
 
-    auto& interface = interfaces_[interfaceName];
+    auto& interface = getInterface(interfaceName);
     InterfaceData::MethodData methodData{ std::move(inputSignature)
                                         , std::move(outputSignature)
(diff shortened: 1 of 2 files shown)


dleeds-cpi:
I'm very happy to see this fix :-)

riuriuriu (approved):
The changes look fine in general. 

One question still: In #177 we also ensured that the `std::map` holds just unique_ptr to the data of which a reference is passed to the underlying sdbus library. Here it would mean to replace: 
```cpp
std::map<InterfaceName, InterfaceData> interfaces_;
```
with
```cpp
std::map<InterfaceName, std::unique_ptr<InterfaceData>> interfaces_;
```

The reasoning was that we don't like to rely on the non-invalidatable references of `std::map`. So someone could later change the container type without knowing of this requirement. Is there a special reason this isn't also done here?

sangelovic:
@riuriuriu No, there is no special reason. Sorry Urs, I re-thought that and I realized even the container changed, the dangerous situation could not happen, becuase the container is always filled first and then used in a non-mutable condition (the items cannot be removed from it, either). It led to slightly simpler code. So I picked this approach. For consistency, I will also do that on the `Proxy` side, in a later MR.

Issue #202, opened 2021-08-20
add `virtual' to interface destructor for generated service/proxy adaptors


AndreyBarmaley:
In an inherited class without a destructor, I get sigabrt:

```
 class MyProxy : public sdbus::ProxyInterface{}

std::unique_ptr<MyProxy> ptr; // <-- SIGABRT
```

sangelovic:
@AndreyBarmaley The code is not compilable. There is no `sdbus::ProxyInterface` class in sdbus-c++. Do you mean `sdbus::ProxyInterfaces` template class? In that case angle brackets with the list of proxy interfaces is missng.  Can you please be more specific and provide more description and full minimal example causing the abort?

sangelovic:
If the problem is addressed in #203 , then I'll close this issue, because #203 is not a solution...

Issue #112, opened 2020-07-10
virtual destructors


tomasz-grobelny:
Is there any reason why sdbus-c++-xml2cpp generates code with non-virtual destructors?

sangelovic:
If you look more closely, the code itself reveals why, since it applies [protected non-virtual destructor idiom](http://www.gotw.ca/publications/mill18.htm).

In short, sdbus-c++-xml2-pp-generated classes are not meant to be used as interface classes by library clients, and specifically of that, they are not meant to be deleted polymorphically. They are meant to be overridden by clients, that's all. Hence, non-virtual protected destructor idiom is applied, and the virtual methods in both proxy and adaptor generated classes are protected.

tomasz-grobelny:
In this case would it make sense to add some pragmas to disable warnings issued by compilers? I am trying to use sdbus-cpp in anbox and came across this issue.

And another thing is that in Message.h and TypeTraits.h there is "struct UnixFd;" forward declaration while definition in Types.h is "class UnixFd". Should this be fixed?

sangelovic:
What kind of warning are you getting there? And which compiler is it that
is giving you a warning on that virtual destructor?

If it's just a non-virtual dtor warning, then pragmas or any code changes
are not a solution at all. The code is valid, safe and idiomatic. What I
would do perhaps is stop using that compiler :) Seriously. I would expect
such warning from a dumb, sub-par linter tool (maybe it's this one that is
giving you this warning?), but surely not from a (mature) compiler.

The latter - mixing of class and struct keyword - is a typo. Thanks for
pointing to it, and please would you submit a pull request? Thaaanks.

tomasz-grobelny:
Ah, my bad. The warning I am getting is from sdbus::AdaptorInterfaces class which is base for my class implementing dbus interface (according to tutorial). And it seems to be correctly generated by gcc 9.3.0 with -Wnon-virtual-dtor command line option.

As for typo - I will when I find some time.

tomasz-grobelny:
Here is the excerpt from gcc output:
```
/home/tomek/anbox/external/sdbus-cpp/include/sdbus-c++/AdaptorInterfaces.h: In instantiation of 'class sdbus::AdaptorInterfaces<org::anbox::ApplicationManager_adaptor>':
/home/tomek/anbox/src/anbox/dbus/ApplicationManagerServer.h:6:48:   required from here
[...]

PR #203, opened 2021-09-14
Generate adaptors and proxies with virtual destructor

Oipo:
The generated classes, when inherited by a base class a la `class MyObject : public sdbus::AdaptorInterfaces<my::ns::MyAdaptor>`, the generated adaptor/proxy does not have a virtual destructor. This causes issues when storing the created MyObject as a MyAdaptor, where upon destruction the destructor of MyObject is never called.

### DIFF
--- a/tools/xml2cpp-codegen/AdaptorGenerator.cpp
+++ b/tools/xml2cpp-codegen/AdaptorGenerator.cpp
@@ -131,7 +131,7 @@ std::string AdaptorGenerator::processInterface(Node& interface) const
                        << propertyRegistration
          << tab << "}" << endl << endl;
 
-    body << tab << "~" << className << "() = default;" << endl << endl;
+    body << tab << "virtual ~" << className << "() = default;" << endl << endl;
 
     if (!signalMethods.empty())
     {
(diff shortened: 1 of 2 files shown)


Oipo:
Fixes https://github.com/Kistler-Group/sdbus-cpp/issues/202

sangelovic:
@Oipo 
Short answer: The destructor is meant to be non-virtual. By design.
Long answer: You cannot store MyObject as a MyAdaptor, because this leads to slicing, and that's not what you want (anyway, generated adaptor/proxy class are anyway mostly abstract classes, so the compiler will not allow that). I assume now you mean storing MyObject pointer as a MyAdaptor pointer, and then calling its destructor. You are now allowed to this, either. The "destructors are either public and virtual or protected and non-virtual" idiom is applied here. The destructor of the generated proxy/adaptor classes is protected. The compiler protects you and won't allow you to shoot your feet. This is consciously designed so. Your adaptor/proxy classes are not ought to be manipulated through their sdbus-c++ generated proxy/adaptor base classes. If you have static analysis issues, simply use `final` keyword in your adaptor/proxy classes.

See https://github.com/Kistler-Group/sdbus-cpp/issues/112 for more discussion.

PR #217, opened 2021-11-13
codegen: Support chrono literal timeout by ProxyGenerator

marcullo:
Allow to use human readable [chrono literals][1] to specify method call
timeout. The change is backward compatbile - if no unit is provided,
the fallback is "us".

Example:
```xml
<annotation name="org.freedesktop.DBus.Method.Timeout" value="500ms"/>
```

[1]: https://en.cppreference.com/w/cpp/symbol_index/chrono_literals

### DIFF
--- a/docs/using-sdbus-c++.md
+++ b/docs/using-sdbus-c++.md
@@ -1156,6 +1156,27 @@ So in the specific example above, the stub generator will generate a `Concatenat
 
 For a real example of a client-side asynchronous D-Bus method, please look at sdbus-c++ [stress tests](/tests/stresstests).
 
+## Method call timeout
+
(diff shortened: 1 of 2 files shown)


docs/using-sdbus-c++.md
```diff
@@ -1156,6 +1156,27 @@ So in the specific example above, the stub generator will generate a `Concatenat
 
 For a real example of a client-side asynchronous D-Bus method, please look at sdbus-c++ [stress tests](/tests/stresstests).
 
+## Method call timeout
+
+Annotate the element with `org.freedesktop.DBus.Method.Async` in order to specify the timeout value for the method call. The value should be a number of microseconds or number with duration literal (`us`/`ms`/`s`/`min`).
```
sangelovic, line 1161: Probably a typo: `org.freedesktop.DBus.Method.Async` -> `org.freedesktop.DBus.Method.Timeout`

sangelovic (changes requested):
Thanks for the contribution. See one comment of mine above (I don't have permissions to do edits in your branch...)

sangelovic (approved):

PR #223, opened 2021-11-28
Fix #214: Add means to unregister signal handler

### LINKED ISSUE #214: Unregister signal handler - Client Side
The client can register a call back handler for any signal but he couldn't unregister it without unregistering everything!
Is there a reason for the current interface design?
I have an easy work-around to provide this feature, my proposal goes as follows:

1. Add a **virtual** method to the `IProxy` interface:
`virtual void unregisterSignalHandler( const std::string& interfaceName, const std::string& signalName ) = 0;`

2. Implement it in the `Proxy` class as removing the relevant entry from the `signals_` map of the given interface.
3. Add a `clear()` method to the `SignalSubscriber` class that call the unregister method.

**Use case:**
`auto subscriber = m_proxy->uponSignal("some_signal").onInterface("some_interface");`
`subscriber.call(some_handler); // register handler`
`// ........`
`// ........`
`subscriber.clear(); // unregister handler`

If my proposal looks good and doesn't break any thing I can create a pull request for this issue.

osamaghanem:
Summary:
I implemented your second approach as follows:
- Add `void IProxy::unregisterSignalHandler(const std::string& interfaceName, const std::string& signalName)` method.
- Add new auxiliary class `SignalUnsubscriber`.
- Add `SignalUnsubscriber IProxy::muteSignal(const std::string& signalName)` method.
- Extend **integrationtests/DBusSignalsTests** by adding 4 testcases for unregistring signal's handler.

### DIFF
--- a/src/Proxy.cpp
+++ b/src/Proxy.cpp
@@ -167,6 +167,16 @@ void Proxy::registerSignalHandler( const std::string& interfaceName
     SDBUS_THROW_ERROR_IF(!inserted, "Failed to register signal handler: handler already exists", EINVAL);
 }
 
+void Proxy::unregisterSignalHandler( const std::string& interfaceName
+                                   , const std::string& signalName )
+{
+    auto& interface = interfaces_[interfaceName];
+
+    auto removeResult = interface.signals_.erase(signalName);
+
+    SDBUS_THROW_ERROR_IF(removeResult == 0, "Failed to unregister signal handler: handler not exists", EINVAL);
+}
+
 void Proxy::finishRegistration()
 {
     registerSignalHandlers(*connection_);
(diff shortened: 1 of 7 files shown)


sangelovic:
Thanks @osamaghanem

src/Proxy.cpp
```diff
@@ -167,6 +167,16 @@ void Proxy::registerSignalHandler( const std::string& interfaceName
     SDBUS_THROW_ERROR_IF(!inserted, "Failed to register signal handler: handler already exists", EINVAL);
 }
 
+void Proxy::unregisterSignalHandler( const std::string& interfaceName
+                                   , const std::string& signalName )
+{
+    auto& interface = interfaces_[interfaceName];
```
dleeds-cpi, line 173: Recommend using `at()` or `find()` here, since `unregisterSignalHandler()` should never add a new entry to `interfaces_`.

sangelovic: Good point. I've created a follow-up MR: https://github.com/Kistler-Group/sdbus-cpp/pull/226

Issue #231, opened 2022-01-10
Potential undefined behaviour in creation of sdbus::Error


BenKaufmann:
At least ` MethodCall::sendWithReply()` and `Proxy::sdbus_async_reply_handler()` create an sdbus::Error object from an `sd_bus_error` object **without** ensuring that the contained **optional** `message` is set (i.e. != `nullptr`). Hence, if the message field is not set, the code tries to create a `std::string` from a `nullptr`, which has undefined behaviour.

Note that e.g. `sd_bus_error_set` **explicitly** allows using `nullptr` to mean "no message".

One simple fix would be to add a second constructor overload: 
```cpp
Error(const std::string& name, const char* message)
  : Error(name, std::string(message ? message :  ""))
{
}
```

sangelovic:
@BenKaufmann Good point. Would you file a PR here, too?

PR #233, opened 2022-01-10
Fix potential UB in creation of sdbus::Error.

BenKaufmann:
See https://github.com/Kistler-Group/sdbus-cpp/issues/231

### DIFF
--- a/include/sdbus-c++/Error.h
+++ b/include/sdbus-c++/Error.h
@@ -43,6 +43,11 @@ namespace sdbus {
         : public std::runtime_error
     {
     public:
+        explicit Error(const std::string& name, const char* message = nullptr)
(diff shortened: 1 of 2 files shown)


include/sdbus-c++/Error.h
```diff
@@ -43,6 +43,11 @@ namespace sdbus {
         : public std::runtime_error
     {
     public:
+        explicit Error(const std::string& name, const char* message = nullptr)
+            : Error(name, std::string(message ? message : ""))
```
sangelovic, line 47: Message parameter with default nullptr value is exactly what I thought about when I saw your commit in your repo. Glad that you've added it (and the tests too)

sangelovic (approved):

PR #242, opened 2022-02-10
refactor: add validity checks for names and paths

### LINKED ISSUE #144: API should check for invalid service, object or interface names
> One of the major sdbus-c++ design goals is to make the sdbus-c++ API easy to use correctly, and hard to use incorrectly.

I hunted down a bug today with this little error message:

      what():  [org.freedesktop.DBus.Error.InvalidArgs] Failed to register object vtable (Invalid argument)

In the end, it was a invalid object-path, namly `/com/corp-domain/service`

It was actually my first self written sdbus server project, so I didn't know that the hyphen is not allowed.
It would be nice, if the API checks all the names first and throws an error if anything is invalid.

sangelovic:
Fixes #144 

/cc @hellow554 @Oxymoron79

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -27,6 +27,7 @@
 #include "Connection.h"
 #include "SdBus.h"
 #include "MessageUtils.h"
+#include "Utils.h"
 #include <sdbus-c++/Message.h>
 #include <sdbus-c++/Error.h>
 #include "ScopeGuard.h"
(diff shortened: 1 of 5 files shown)


hellow554:
Thanks. Looks good to me. 👍

src/Connection.cpp
```diff
@@ -71,6 +72,7 @@ Connection::~Connection()
 
 void Connection::requestName(const std::string& name)
 {
+    SDBUS_CHECK_SERVICE_NAME(name);
```
Oxymoron79, line 75: Extreme nitpick, since all other checks have a empty line after the checks.
```suggestion
    SDBUS_CHECK_SERVICE_NAME(name);
    
```

Oxymoron79 (approved):

PR #246, opened 2022-02-20
fix: missing object enumeration for GetManagedObjects

akemnade:
If an ObjectManager is added to an object, GetManagedObjects
returns an empty list due to missing enumerator.
Add the enumerator, for now just handling the object path itself
without children


akemnade:
fixes #247

sangelovic:
I'll look at this soon

akemnade:
any chance to get this reviewed?

akemnade:
and why this was marked as draft without mentioning anything which should be improved/fixed before it can be considered ready?

dleeds-cpi:
The default node enumerator supplied by sd-bus should be sufficient for this library's purposes. A custom enumerator is only needed for scenarios where we want the Object Manager to return object paths that are not registered via a v-table.

PR #271, opened 2022-07-07
Add a wrapper for sd-event integration

hadrianw:
As discussed here: https://github.com/Kistler-Group/sdbus-cpp/discussions/262

### DIFF
--- a/include/sdbus-c++/IConnection.h
+++ b/include/sdbus-c++/IConnection.h
@@ -34,6 +34,8 @@
 #include <cstdint>
 #include <optional>
 
+struct sd_event;
+
 namespace sdbus {
 
     /********************************************//**
@@ -132,6 +134,25 @@ namespace sdbus {
          * @throws sdbus::Error in case of failure
          */
         virtual std::string getUniqueName() const = 0;
+
+        /*!
+         * @brief Attach the bus connection to an sd-event event loop
+         *
+         * @throws sdbus::Error in case of failure
+         */
+        virtual void attachSdEventLoop(sd_event *ev, int priority = 0) = 0;
+
+        /*!
+         * @brief Detach the bus connection from an sd-event event loop
+         *
+         * @throws sdbus::Error in case of failure
+         */
+        virtual void detachSdEventLoop() = 0;
+
+        /*!
+         * @brief Get current sd-event event loop for the bus connection
+         */
+        virtual sd_event *getSdEventLoop() = 0;
 
         /*!
          * @brief Enters I/O event loop on this bus connection
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -92,6 +92,24 @@ std::string Connection::getUniqueName() const
     return unique;
 }
 
+void Connection::attachSdEventLoop(sd_event *ev, int priority)
+{
+    auto r = iface_->sd_bus_attach_event(bus_.get(), ev, priority);
+    SDBUS_THROW_ERROR_IF(r < 0, "Failed to attach to event loop", -r);
+}
+
+void Connection::detachSdEventLoop()
+{
+    auto r = iface_->sd_bus_detach_event(bus_.get());
+    SDBUS_THROW_ERROR_IF(r < 0, "Failed to detach from event loop", -r);
+}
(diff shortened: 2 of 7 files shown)


include/sdbus-c++/IConnection.h
```diff
@@ -133,6 +135,25 @@ namespace sdbus {
          */
         virtual std::string getUniqueName() const = 0;
 
+        /*!
+         * @brief Attach the bus connection to an sd-event event loop
+         *
+         * @throws sdbus::Error in case of failure
+         */
+        virtual void attachEventLoop(sd_event *ev, int priority) = 0;
```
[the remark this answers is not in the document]
hadrianw: Done

include/sdbus-c++/IConnection.h
```diff
@@ -133,6 +135,25 @@ namespace sdbus {
          */
         virtual std::string getUniqueName() const = 0;
 
+        /*!
+         * @brief Attach the bus connection to an sd-event event loop
+         *
+         * @throws sdbus::Error in case of failure
+         */
+        virtual void attachEventLoop(sd_event *ev, int priority = 0) = 0;
```
dleeds-cpi, line 143: Suggestion. Because there are already `enterEventLoop()` and `enterEventLoopAsync()` functions defined, and they are not compatible with these new event loop related calls, it might be helpful to be more explicit with the naming. E.g. `attachSdEventLoop()`, `detachSdEventLoop()`, etc.

Another thing that came to mind, was the possibility of using the `sd-event` loop as the built-in event loop to make the integration consistent. However, I am not aware of any hooks `sd-event` provides to notify an application when the event loop is about to start and end processing. Thus, there would be no way to know when to lock `sdbusMutex_` and preserve the thread-safety of the bus state. So, unless there is something I am missing, that is not possible with the current `libsystemd` API.

hadrianw: I will change those names.

Regarding integration of sd-event as the event loop I think it should be doable by not using sd_event_loop. Take a look at https://www.freedesktop.org/software/systemd/man/sd_event_run.html you could go a step or two lower level. Or maybe I'm not fully understanding how it should work.

hadrianw: Done

sangelovic:
@hadrianw Thanks a lot for your effort, and @dleeds-cpi for your timely review. The PR looks good per se, *but*..,

I looked now more deeply into sd-event (I wasn't familiar with it) and its integration in sd-bus. And I see we can't take the path of simply delegating `attachSdEventLoop` to `sd_bus_attach_event`. As @dleeds-cpi already mentioned in a comment above, this would create direct relationship between sd-event and sd-bus and would bypass sdbus-c++. sdbus-c++ is designed around multi-threaded use, and this would mean bypassing its thread safety mechanisms (like `sdbusMutex_`), which would lead to the realm of undefined behavior. Even if sd-event had (AFAICS it doesn't) some hooks and and lower-level API, we can't rely on the client to use them.

I think the only option is implementing `attachSdEventLoop`/`detachSdEventLoop` in sdbus-c++. It could be written along the lines of internal sd-bus implementation, but would be based on fd, events and timeouts provided by public sd-bus API.

What do you think? Do you have other ideas? Would you volunteer to implement that in this PR?

include/sdbus-c++/IConnection.h
```diff
@@ -133,6 +135,25 @@ namespace sdbus {
          */
         virtual std::string getUniqueName() const = 0;
 
+        /*!
+         * @brief Attach the bus connection to an sd-event event loop
+         *
+         * @throws sdbus::Error in case of failure
+         */
+        virtual void attachSdEventLoop(sd_event *ev, int priority = 0) = 0;
```
sangelovic, line 143: These new virtual methods shall be moved all the way down in the class, just after `addMatch()` and before deprecated `enterProcessingLoop()`, to not break ABI compatibility, and to not have to create a new major version...

sangelovic:
sd-event integration was added in #310. It will be released with new major release of sdbus-c++ v2.0 in the coming months.

PR #273, opened 2022-07-14
feat: add support for session bus connection at custom address

alivenets:
The new function helper `createSessionBusConnectionWithAddress` allows
to create connection to session bus with custom address.

Signed-off-by: Alexander Livenets <a.livenets@gmail.com>


sangelovic:
Thanks @alivenets

PR #279, opened 2022-08-02
feat: add support for proxy with no event loop thread

sangelovic:
This adds a possibility to create a proxy which has its own D-Bus connection to a bus, but which doesn't run an event loop thread for it internally. Such proxies are "light-weight", meant to be created, used (for some classic D-Bus calls), and discarded.


sangelovic:
I will update the docs (tutorial), but otherwise it's ready for review...

sangelovic:
@Oxymoron79 @lubo-svk I was removed from Kistler-Group group, and lost maintainer privileges for sdbus-c++. Do you know what happened? Why?

lubo-svk:
Hello, I don't know. I will look on it tomorrow and let you know.

sangelovic:
@lubo-svk OK, thanks :)

lubo-svk:
There should be an invite for you. One colleague did a cleanup and
accidentally removed you.

PR #280, opened 2022-08-03
refactor: use pseudo D-Bus connection for plain messages

### LINKED ISSUE #277: Is there a way to create `PlainMessage`s without a bus connection?
[sorry if this is more of a support request than a regular issue, i guess you could phrase it as "please document" instead though]

During unit-testing of my code i've come across the need to create `sdbus::PlainMessage`s as test input. I didn't find a way to create them so they are valid/usable without having an actual D-Bus daemon running (i.e. `operator>>` would fail without D-Bus connection). Is there something i'm missing here, or is this not a supported use-case?

sangelovic:
This fixes #277 and addresses @dleeds-cpi's proposal that is part of discussion #169.

To create a plain D-Bus messages (used to implement the concept of `Variant`s, for example), we need a D-Bus connection. However, with current sd-bus API and its implementation, it's possible to optimize and create a "pseudo" connection which does not really connect to and thus does not need D-Bus daemon, yet remains in a state that is sufficient to create local D-Bus messages and work with them.

This is a bit of a hack, but

* it enables us and our clients to write unit tests around plain messages and `Variant`s that fulfill FIRST principles, and are independent from presence of D-Bus daemon.
* leads to faster and more light-weight code when creating plain D-Bus messages/`Variant`s.

@dleeds-cpi would you please review? :)

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -70,6 +70,13 @@ Connection::Connection(std::unique_ptr<ISdBus>&& interface, remote_system_bus_t,
 {
 }
 
+Connection::Connection(std::unique_ptr<ISdBus>&& interface, pseudo_bus_t)
+    : iface_(std::move(interface))
+    , bus_(openPseudoBus())
+{
+    assert(iface_ != nullptr);
+}
+
 Connection::~Connection()
 {
     Connection::leaveEventLoop();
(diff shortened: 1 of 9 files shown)


dleeds-cpi:
@sangelovic looks okay to me. I hate to incorporate this hack`, but unfortunately, `sd-bus` seems quite attached to pairing a message with a particular bus connection, even though messages are rarely sent without the caller specifying the bus.

PR #283, opened 2022-08-20
Don't include libsystemd as pkg-config dependency when statically linked

chewi:
This works whether `BUILD_LIBSYSTEMD` is `ON` or the separate build of libsystemd picked up by pkg-config is static. The Gentoo Linux package requires the latter.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,6 +210,11 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/cmake/sdbus-c++-config.cmake
         DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/sdbus-c++
         COMPONENT dev)
 
+if(BUILD_SHARED_LIBS AND (BUILD_LIBSYSTEMD OR Systemd_LINK_LIBRARIES MATCHES "/libsystemd\.a(;|$)"))
+    set(PKGCONFIG_REQS ".private")
+else()
+    set(PKGCONFIG_REQS "")
+endif()
 configure_file(pkgconfig/sdbus-c++.pc.in pkgconfig/sdbus-c++.pc @ONLY)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/pkgconfig/sdbus-c++.pc
         DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig COMPONENT dev)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -210,6 +210,11 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/cmake/sdbus-c++-config.cmake
         DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/sdbus-c++
         COMPONENT dev)
 
+if(BUILD_LIBSYSTEMD OR pkgcfg_lib_Systemd_systemd MATCHES "\.a$")
```
sangelovic, line 213: I could not find any documentation on variables named according to the `pkgcfg_lib_*` pattern. Could you point me to some? Or aren't these internal variables?

In the latter case, wouldn't it be better to use documented variables -- i.e., `Systemd_LINK_LIBRARIES`, for example, instead of the internal `pkgcfg_lib_Systemd_systemd`?

sangelovic (changes requested):
Thanks for the contribution, I have one comment to your PR...

CMakeLists.txt
chewi: Good call. I'm no CMake expert, so I found that variable by diffing the build trees of a static and dynamic build. That was the only difference. I'm still surprised I overlooked the documented variable though. It's a `LIST`, but I was still able to use `MATCHES` against it.

sangelovic:
@chewi Thanks for fixing. One additional idea that popped in my mind... What if sdbus-c++ itself is built and distributed as a static lib? In that case we do need to have libsystemd as a `requires` in pkg-config file of sdbus-c++, even if libsystemd is distributed as static library, don't we? What do you say?

chewi:
Another good point! It's a slightly different case, but this reminds me of similar fixes I made to curl and SDL2. I've now changed the way it works. It's a bit confusing, but hopefully the additional text in the commit message explains it clearly enough:

> This change also handles `BUILD_SHARED_LIBS` being `OFF`. In this case, the libsystemd dependency does need to be included, even though it is static.
> 
> CMake does not allow you to build shared and static libraries simultaneously (with a single `add_library` call), but this change declares the libsystemd dependency as `Requires.private` rather than omitting it entirely. This is in case someone builds and installs both a static and shared build and then calls `pkg-config --static sdbus-c++`. The static build needs to be installed first so that the pkg-config file from the shared build takes precedence.

sangelovic:
@chewi Superb, thanks! :-)

PR #287, opened 2022-09-04
Force event loop to re-enter processing to handle queued messages

zaporozhets:
Hi everyone.

Unfortunately, an external event loop can not handle some pending events(signals for example). This issue happens when library calls `sd_bus_call`. Function `sd_bus_call` reads all available data from sdbus file descriptor. As the result, the polling mechanism in external event loop can not detect changes on sdbus file descriptor.

Proposed fix forces event loop to re-enter processing to handle queued messages.

Cheers,
Taras

### DIFF
--- a/include/sdbus-c++/IConnection.h
+++ b/include/sdbus-c++/IConnection.h
@@ -67,10 +67,6 @@ namespace sdbus {
             /*!
              * The read fd to be monitored by the event loop.
              */
-            int event_fd;
-            /*!
-             * The read fd to be monitored by the event loop.
-             */
             int fd;
             /*!
              * The events to use for poll(2) alongside fd.
@@ -81,6 +77,11 @@ namespace sdbus {
              * Absolute timeout value in micro seconds and based of CLOCK_MONOTONIC.
              */
             uint64_t timeout_usec;
+
+            /*!
+             * The read fd to be monitored by the event loop.
+             */
+            int event_fd;
 
             /*!
              * Get the event poll timeout.
--- a/include/sdbus-c++/IConnection.h
+++ b/include/sdbus-c++/IConnection.h
@@ -67,6 +67,10 @@ namespace sdbus {
             /*!
              * The read fd to be monitored by the event loop.
              */
+            int event_fd;
+            /*!
+             * The read fd to be monitored by the event loop.
+             */
             int fd;
             /*!
              * The events to use for poll(2) alongside fd.
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -146,7 +146,7 @@ Connection::PollData Connection::getEventLoopPollData() const
     auto r = iface_->sd_bus_get_poll_data(bus_.get(), &pollData);
     SDBUS_THROW_ERROR_IF(r < 0, "Failed to get bus poll data", -r);
 
-    return {eventFd_.fd, pollData.fd, pollData.events, pollData.timeout_usec};
+    return {pollData.fd, pollData.events, pollData.timeout_usec, eventFd_.fd};
 }
 
 const ISdBus& Connection::getSdBusInterface() const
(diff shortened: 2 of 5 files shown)


sangelovic:
@zaporozhets Hello, can you please provide more details on your use case? Does this involve multiple threads, i.e. `sd_bus_call` takes place in one thread while external polling is taking place in another thread? In such a case, incoming messages should wake up all polling threads, the first and the second one. The second one (external event loop) will just wait in `processPendingRequest()` while the first one finishes its call, and the the second one will proceed with processing the queued messages.
Or is your situation single-threaded, i.e. you do `sd_bus_call` in a thread and then proceed to polling in that thread? Yes, in such case the `sd_bus_call` will drain the incoming messages, and the next call to poll won't wake up the thread to process the queued messages. The solution here would be that you call `processPendingRequest()` in your event loop before entering poll. Or is you situation yet different?

zaporozhets:
Hi @sangelovic.
Thank you. Yes, I'm running in single-threaded environment. And as you mentioned, `sd_bus_call` drains the incoming messages. Your recommendation regarding manually calling `processPendingRequest` will work here. But would be nice to hide this under the hood of the library. What do you think? 

I also have prepared an example to demonstrate this problem and potential fix for it. The following repo contains two branches(master and fix/updated_event_loop) with an example before and after applying the fix.

zaporozhets:
@sangelovic, is there anything that I can do to improve this MR?

sangelovic:
@zaporozhets Hi, thanks for your patience :) I appreciate your effort to contribute, and the MR is not bad per se. I did look at your example yesterday and I realized the cause is elsewhere. The cause is that in your example, poll timeout is not updated before each poll. Underlying sd-bus library is designed in such a way that that poll timeouts shall be queried and used in the poll in case of external loops. Citing from [`sd_bus_get_timeout` man page](https://www.freedesktop.org/software/systemd/man/sd_bus_get_timeout.html#):

> These three functions are useful to hook up a bus connection object with an external or manual event loop involving poll() or a similar I/O polling call. Before each invocation of the I/O polling call, all three functions should be invoked: the file descriptor returned by sd_bus_get_fd() should be polled for the events indicated by sd_bus_get_events(), and the I/O call should block for that up to the timeout returned by sd_bus_get_timeout(). After each I/O polling call the bus connection needs to process incoming or outgoing data, by invoking [sd_bus_process(3)](https://man.archlinux.org/man/sd_bus_process.3.en).

I took master branch of your example and modified it (tried to push my branch to remote but don't have permissions) to get the timeout (through `IConnection::getEventLoopPollData()` call) after each `process()` call and use that timeout in asio. And the app works like a charm.

So the changes in this MR are AFAICS not necessary.

Perhaps a subsection in the sdbus-c++ tutorial explaining how to well integrate sdbus-c++ into external foreign loops would be handy for users like you?

sangelovic:
@zaporozhets For an example, you can also look directly into how sd-bus integrates with sd-event. The [`prepare_callback()`](https://github.com/systemd/systemd/blob/main/src/libsystemd/sd-bus/sd-bus.c#L3621) function is called before each poll in sd-event. This function always gets current poll events and poll timeout...

In your example, the timeout returned before the poll following the synchronous call is zero, since sd-bus can see that there are messages in the input queue. So the next poll will return immediately, leading to the `process()` method invocation which immediately processes the messages that arrived during the sync call.

sangelovic:
@zaporozhets Do you have any additional thoughts on this, or can we close this MR?

sangelovic:
@zaporozhets This issue has been fixed in #309. Now for both concurrent and single-threaded use of poll and sd_bus_call, if there are extra incoming messages (like signals) drained by the sd_bus_call on the fd, the event loop will be woken up to process those messages.

sangelovic:
The fix will come with new major release - v2.0.0, since it breaks the API, ABI and changes some existing behaviors. It may take some time until this version is released, since more backwards-incompatible improvements are planned to be worked on and packed in that release.

Issue #312, opened 2023-01-31
Potential race condition and memory leak


anthonybrandon:
**Describe the bug**

I think there is a potential race condition and resulting memory leak in `Proxy::callMethod` (the async one).

**To Reproduce**

Call an async method from a different thread than the event loop.

**Expected behavior**

The AsyncCalls::CallData should be freed in sdbus_async_reply_handler.

**Real (buggy) behavior**

sdbus_async_reply_handler is called before the AsyncCalls::CallData can be added to pendingAsyncCalls_.
Afterwards the CallData is added to pendingAsyncCalls_, but then it is never removed.
This results in a memory leak. (calls_ grows every time an async call is made).

**Additional context**

I'm only able to reproduce this on a very slow CPU. Maybe if you add some sleep after message.send() you can reproduce it on a normal machine.
I'm also not 100% sure if making an async call from a different thread than the event loop thread is supposed to be thread safe, but it seems like it should be.

Edit:

I guess technically not a memory leak, since there is still a reference to the CallData, but the calls_ map does grow until you run out of memory.

anthonybrandon:
I was able to get rid of the leak with the following changes, but I'm not sure if this is the correct approach to fix it.

```diff
diff --git a/src/Proxy.cpp b/src/Proxy.cpp
index 7f53ae1..37dc290 100644
--- a/src/Proxy.cpp
+++ b/src/Proxy.cpp
@@ -108,11 +108,9 @@ PendingAsyncCall Proxy::callMethod(const MethodCall& message, async_reply_handle
     auto callData = std::make_shared<AsyncCalls::CallData>(AsyncCalls::CallData{*this, std::move(asyncReplyCallback), {}});
     auto weakData = std::weak_ptr<AsyncCalls::CallData>{callData};
 
+    pendingAsyncCalls_.addCall(callData);
     callData->slot = message.send(callback, callData.get(), timeout);
 
-    auto slotPtr = callData->slot.get();
-    pendingAsyncCalls_.addCall(slotPtr, std::move(callData));
-
     return {weakData};
 }
 
@@ -238,7 +236,6 @@ int Proxy::sdbus_async_reply_handler(sd_bus_message *sdbusMessage, void *userDat
     assert(asyncCallData != nullptr);
     assert(asyncCallData->callback);
     auto& proxy = asyncCallData->proxy;
-    auto slot = asyncCallData->slot.get();
 
[...]

PR #313, opened 2023-02-01
Fix race condition in Proxy::callMethod.

anthonybrandon:
After calling message.send it is possible for the callback to be called before callData is added to pendingAsyncCalls_. This results in callData being added there after the callback is called and never being removed.

Signed-off-by: Anthony Brandon <anthony@amarulasolutions.com>

### DIFF
--- a/src/Proxy.cpp
+++ b/src/Proxy.cpp
@@ -123,10 +123,16 @@ PendingAsyncCall Proxy::callMethod(const MethodCall& message, async_reply_handle
     auto callData = std::make_shared<AsyncCalls::CallData>(AsyncCalls::CallData{*this, std::move(asyncReplyCallback), {}});
     auto weakData = std::weak_ptr<AsyncCalls::CallData>{callData};
 
-    callData->slot = message.send(callback, callData.get(), timeout);
-
-    auto slotPtr = callData->slot.get();
-    pendingAsyncCalls_.addCall(slotPtr, std::move(callData));
+    pendingAsyncCalls_.addCall(callData);
+    try
+    {
+        callData->slot = message.send(callback, callData.get(), timeout);
+    }
+    catch (...)
+    {
+        pendingAsyncCalls_.removeCall(callData.get());
+        throw;
+    }
 
     return {weakData};
 }
@@ -253,7 +259,6 @@ int Proxy::sdbus_async_reply_handler(sd_bus_message *sdbusMessage, void *userDat
     assert(asyncCallData != nullptr);
     assert(asyncCallData->callback);
     auto& proxy = asyncCallData->proxy;
-    auto slot = asyncCallData->slot.get();
 
     // We are removing the CallData item at the complete scope exit, after the callback has been invoked.
     // We can't do it earlier (before callback invocation for example), because CallBack data (slot release)
@@ -261,8 +266,7 @@ int Proxy::sdbus_async_reply_handler(sd_bus_message *sdbusMessage, void *userDat
     SCOPE_EXIT
     {
         // Remove call meta-data if it's a real async call (a sync call done in terms of async has slot == nullptr)
-        if (slot)
-            proxy.pendingAsyncCalls_.removeCall(slot);
+        proxy.pendingAsyncCalls_.removeCall(asyncCallData);
     };
 
     auto message = Message::Factory::create<MethodReply>(sdbusMessage, &proxy.connection_->getSdBusInterface());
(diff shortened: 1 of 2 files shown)


anthonybrandon:
See #312 for the full explanation.

sangelovic:
@anthonybrandon Thanks for contribution, looks good to me in overall.

One thing I don't like, however, is that the side effect of storing the async call struct into the collection *before* the actual sending the call message is that we have to revert it in case the sending throws an exception.

Somewhat less complex and more declarative solution than try-catch clause would IMO be to use scope guard (which is already used in several places across sdbus-c++):

```c++
    pendingAsyncCalls_.addCall(callData);
    SCOPE_EXIT_NAMED(removeAsyncCallOnException){ pendingAsyncCalls_.removeCall(callData.get()); };
    callData->slot = message.send(callback, callData.get(), timeout);
    removeAsyncCallOnException.dismiss();
```

Still, the order is inverted (first pushing into the collection, then the call). As an alternative, we can keep the logical order, and introduce a flag member into `CallData` struct:

```c++
            struct CallData
            {
                Proxy& proxy;
                async_reply_handler callback;
                Slot slot;
                bool finished{false};
            };
```

The flag would tell if the call has finished or not yet. We could then query and set the flag here (I would also refactor `addCall()` to return `void`):

```c++
            void addCall(std::shared_ptr<CallData> asyncCallData)
            {
                std::lock_guard lock(mutex_);
                if (!asyncCallData->finished) // The call may have finished in the mean time
                    calls_.emplace_back(std::move(asyncCallData));
            }

            void removeCall(CallData const* data)
            {
                std::unique_lock lock(mutex_);
                if (auto it = std::find_if(calls_.begin(), calls_.end(), [data](auto const& entry){ return entry.get() == data; }); it != calls_.end())
                {
                    auto callData = std::move(*it);
                    calls_.erase(it);
                    callData->finished = true;
                    lock.unlock();

                    // Releasing call slot pointer acquires global sd-bus mutex. We have to perform the release
                    // out of the `mutex_' critical section here, because if the `removeCall` is called by some
                    // thread and at the same time Proxy's async reply handler (which already holds global sd-bus
                    // mutex) is in progress in a different thread, we get double-mutex deadlock.
                }
            }
```

The synchronization of access to the `finished` member is guaranteed by the `mutex_`, so there is no additional synchronization cost. The extra cost is a little bigger `CallData` struct. But I don't think this is an issue at all, and what we get is more readable code at a higher level of abstraction with logical flow. I'd go for this option.

What do you think?

anthonybrandon:
@sangelovic Overall I think you're right, I prefer some version that doesn't have to add, and then in case of failure remove.
I think your suggestion can work, but I think it needs to be modified a bit. If I've understood correctly, `finished` should be set to true outside of the if condition, since in the case of the race it won't have been added yet, and that is exactly what we need to protect against.

```cpp
            void addCall(std::shared_ptr<CallData> asyncCallData)
            {
                std::lock_guard lock(mutex_);
                if (!asyncCallData->finished) // The call may have finished in the mean time
                    calls_.emplace_back(std::move(asyncCallData));
            }

            void removeCall(CallData const* data)
            {
                std::unique_lock lock(mutex_);
                data->finished = true;
                if (auto it = std::find_if(calls_.begin(), calls_.end(), [data](auto const& entry){ return entry.get() == data; }); it != calls_.end())
                {
                    auto callData = std::move(*it);
                    calls_.erase(it);
                    lock.unlock();

                    // Releasing call slot pointer acquires global sd-bus mutex. We have to perform the release
                    // out of the `mutex_' critical section here, because if the `removeCall` is called by some
                    // thread and at the same time Proxy's async reply handler (which already holds global sd-bus
                    // mutex) is in progress in a different thread, we get double-mutex deadlock.
                }
            }
```

One other question I had is if it's fine to store these in a vector? Previously it was an unordered_map. I'm not sure if there could be some performance impact.

anthonybrandon:
I'll test this and update my PR.

sangelovic:
@anthonybrandon Good point, `data->finished = true;` needs to be done outside of the if condition. Regarding vector vs. map, it depends. For relatively small number of pending async calls, the vector outperforms the map. With larger number, the map may start to be faster than the vector. But it pretty much depends on where in the vector we remove the call item, because the remaining elements must all be shifted to the left. Fortunately, `shared_ptr`'s move constructor is noexcept so the vector shall choose move operations, which are very cheap (much cheaper than copy operations involving reference count changes). It may help if we chose `std::deque` over `std::vector`, in case there is a number of pending async calls and if they finish in the same order that they started, then in case of vector all elements would also need to be shifted, but in case of deque removing the first element doesn't cause shifting of all the reamining elements.

sangelovic:
@anthonybrandon OK, so we agree on the `finished` flag-based solution. And I would slightly prefer deque over vector, but I'll leave it up to you. If you update the MR I can merge it already today.

PR #325, opened 2023-05-15
Prevent installing a time event, when there's no need for a timeout

### LINKED ISSUE #324: New Integration of sd-event loop
On the `release/v2.0` branch there is an ongoing work to integrate this library with the sd-event loop.

Currently, I'm not able to have it working, because as soon as I try to run the event loop, it errors out with `-EINVAL`, because of one of the timerfd syscalls  returns `EINVAL`

(strace of my program):
```
timerfd_settime(12, TFD_TIMER_ABSTIME, {it_interval={tv_sec=0, tv_nsec=0}, it_value={tv_sec=-1, tv_nsec=4294967295}}, NULL) = -1 EINVAL (Invalid argument)
```

After investigating the `attachSdEventLoop` function a bit, I wonder why you don't use the `sd_bus_attach_event` at all? Wouldn't that be much simpler?

hellow554:
Fixes #324

All tests are passing on both x86_64 and arm:

```
# /usr/lib/sdbus-c++/tests/sdbus-c++-unit-tests --gtest_brief=1
[==========] 105 tests from 43 test suites ran. (14 ms total)
[  PASSED  ] 105 tests.
# /usr/lib/sdbus-c++/tests/sdbus-c++-integration-tests --gtest_brief=1
[==========] 149 tests from 10 test suites ran. (24349 ms total)
[  PASSED  ] 149 tests.
```


CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 
 cmake_minimum_required(VERSION 3.13)
 
-project(sdbus-c++ VERSION 1.2.0 LANGUAGES C CXX)
+project(sdbus-c++ VERSION 2.0.0 LANGUAGES C CXX)
```
hellow554, line 7: I did this, because the library still gets created as `libsdbus-c++.so.1.2.0` which is not correct

hellow554:
Meh... with this patch I get a 100% CPU utilization, because the timer now triggers indefinitely:

```
timerfd_settime(14, TFD_TIMER_ABSTIME, {it_interval={tv_sec=0, tv_nsec=0}, it_value={tv_sec=0, tv_nsec=1}}, NULL) = 0
epoll_wait(3, [{events=EPOLLIN, data={u32=4704016, u64=4704016}}], 12, -1) = 1
clock_gettime64(CLOCK_REALTIME, {tv_sec=1637344918, tv_nsec=304498532}) = 0
clock_gettime64(CLOCK_MONOTONIC, {tv_sec=2556, tv_nsec=542450772}) = 0
clock_gettime64(CLOCK_BOOTTIME, {tv_sec=2556, tv_nsec=542974452}) = 0
read(14, "\1\0\0\0\0\0\0\0", 8)         = 8
recvmsg(8, {msg_namelen=0}, MSG_DONTWAIT|MSG_CMSG_CLOEXEC) = -1 EAGAIN (Resource temporarily unavailable)
read(10, 0xbe9226c8, 8)                 = -1 EAGAIN (Resource temporarily unavailable)
gettid()                                = 715
timerfd_settime(14, TFD_TIMER_ABSTIME, {it_interval={tv_sec=0, tv_nsec=0}, it_value={tv_sec=0, tv_nsec=1}}, NULL) = 0
epoll_wait(3, [{events=EPOLLIN, data={u32=4704016, u64=4704016}}], 12, -1) = 1
clock_gettime64(CLOCK_REALTIME, {tv_sec=1637344918, tv_nsec=308156049}) = 0
clock_gettime64(CLOCK_MONOTONIC, {tv_sec=2556, tv_nsec=546113448}) = 0
clock_gettime64(CLOCK_BOOTTIME, {tv_sec=2556, tv_nsec=546480480}) = 0
read(14, "\1\0\0\0\0\0\0\0", 8)         = 8
recvmsg(8, {msg_namelen=0}, MSG_DONTWAIT|MSG_CMSG_CLOEXEC) = -1 EAGAIN (Resource temporarily unavailable)
read(10, 0xbe9226c8, 8)                 = -1 EAGAIN (Resource temporarily unavailable)
gettid()                                = 715
timerfd_settime(14, TFD_TIMER_ABSTIME, {it_interval={tv_sec=0, tv_nsec=0}, it_value={tv_sec=0, tv_nsec=1}}, NULL) = 0
[...]
```

sangelovic:
@hellow554 

> Meh... with this patch I get a 100% CPU utilization, because the timer now triggers indefinitely:

I think that's because we are missing the other piece of puzzle here: https://github.com/systemd/systemd/blob/c06d2e440c07756391c527814611a7f5633a12d9/src/libsystemd/sd-bus/sd-bus.c#L3702. sd-bus does turn off the timer in case r == 0. In other cases it sets one-shot timer.

We shall do the same then.

hellow554:
Ah, I forgot [the second part](https://github.com/systemd/systemd/blob/c06d2e440c07756391c527814611a7f5633a12d9/src/libsystemd/sd-bus/sd-bus.c#L3702):

```c
sd_event_source_set_enabled(bus->time_event_source, r > 0 ? SD_EVENT_ONESHOT : SD_EVENT_OFF);
```

When I adjust it like that, there are some errors :/

```
/usr/src/debug/sdbus-c++/2.0.0.0-r0/git/tests/integrationtests/DBusMethodsTests.cpp:59: Failure
Expected: this->m_proxy->noArgNoReturn() doesn't throw an exception.
  Actual: it throws sdbus::Error with description "[org.freedesktop.DBus.Error.Timeout] Connection timed out".
```

sangelovic:
1. Does the test report the timeout immediately or after circa 25 seconds of being stalled?
2. Can you try `sd_event_source_set_enabled(sdTimeEventSource, set_new_timer ? SD_EVENT_ONESHOT);` version? Does that work without reporting the timeout error in the tests? If so, then we are sure it's the issue of using `SD_EVENT_OFF`, not the issue of `SD_EVENT_ONESHOT`.

sangelovic:
3. Would you try to do exactly the version that is in sd-bus? I.e., use the return value from `sd_bus_get_events()` to decide what to do with the timer. We are now deciding upon the timeout value, but it may be that such a max timeout value is also returned in other "normal" cases ( where r > 0), where we do want to set the timer on.

hellow554:
1. yes:
```
/usr/src/debug/sdbus-c++/2.0.0.0-r0/git/tests/integrationtests/DBusMethodsTests.cpp:59: Failure
Expected: this->m_proxy->noArgNoReturn() doesn't throw an exception.
  Actual: it throws sdbus::Error with description "[org.freedesktop.DBus.Error.Timeout] Connection timed out".
[  FAILED  ] SdbusTestObject/1.CallsEmptyMethodSuccesfully, where TypeParam = sdbus::test::SdEventLoop (25030 ms)
```

2. you mean something like this?

```cxx
if (set_new_timer) {
    r = sd_event_source_set_time(sdTimeEventSource, static_cast<uint64_t>(sdbusPollData.timeout.count()));
    SDBUS_THROW_ERROR_IF(r < 0, "Failed to set timeout for time event source", -r);
    r = sd_event_source_set_enabled(sdTimeEventSource, SD_EVENT_ONESHOT);
    SDBUS_THROW_ERROR_IF(r < 0, "Failed to enable time event source", -r);
} 
```

if so, same error

hellow554:
Can you please confirm that the test cases are correct?

sangelovic:
I don't see why test cases shall not be correct. I changed on my native Linux this `r = sd_event_source_set_enabled(sdTimeEventSource, SD_EVENT_ON);` to this: `r = sd_event_source_set_enabled(sdTimeEventSource, SD_EVENT_ONESHOT);` and it works. You have timeouts on your ARM device or in your native system, or both?

sangelovic:
The failing test abstracts away from the underlying event handling, and does a very simple thing -- a simple D-Bus method call. If that fails, we have an issue in the code.

hellow554:
On my x86_64 machine everything works, but not on my arm device.
I'm not sure why there's a distinction between these two platforms, and I can't find any reason why it shouldn't work.

That's a bummer...

sangelovic:
@hellow554 

I found the issue. You were quite close.

First, there is one side thing of interest -- the max timeout value is converted by sdbus-c++ from `uint64_t` (18446744073709551615) to `int64_t` max value (9223372036854775807) (because that's how `std::chrono::microseconds` is represented), and then passed to `sd_event_source_set_time`. If we use the original `uint64_t` max, then voila -- it works on 32-bit ARM. Somehow the underlying function(s) can handle this value probably as a sentinel value, but the other value they can't. But only on 32-bit.

But in the end, we don't have to care about matching the max value. We can keep `int64_t` max there, we just have to properly enable/disable the timer depending on that value. So we just modify [line 385](https://github.com/Kistler-Group/sdbus-cpp/blob/release/v2.0/src/Connection.cpp#L385) like this:

```
    r = sd_event_source_set_enabled(sdTimeEventSource, sdbusPollData.timeout != sdbusPollData.timeout.max() ? SD_EVENT_ONESHOT : SD_EVENT_OFF);
```

This works on my x64 native as well as ARM32. Can you verify on your machines and let me know?

sangelovic:
If it works on your machines, please commit my suggestion (if you agree with it) into this PR and I'll merge it.

sangelovic:
And sorry for replying late today, I worked on many things simultaneously and left this one to the end of my working day, you know it :)

hellow554:
> 503  Service Unavailable [IP: 52.252.75.106 80]

I'll try later again by a force push

sangelovic:
Do you have a specific reason to bump default googletest version up to 1.13.0 in this PR? Does the timer fix depend on it?

sangelovic:
I would suggest to do the sdbus-c++ major version bump in a separate PR. Since we use squashing and rebasing here for clean history, that will yield a separate, self-contained commit instead of one commits mixing two or three unrelated things...

sangelovic:
Nevermind about my last comment on splitting to separate MR's. As long as it's split in your commits in this PR, we'll do just do rebasing and leave out squashing. But we use [conventional changelog](https://www.conventionalcommits.org/en/v1.0.0/) style. Could you please rename your commits to match that style?

sangelovic:
> ```
> [30/37] Building CXX object _deps/googletest-build/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o
> FAILED: _deps/googletest-build/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o
>...
> ```

Where does this come from? Your local build? Or CI pipeline here?

In case of former, the solution is not to build with `-Werror`, since that may be fragile for external libraries like googletest.

In case of latter, I have prepared a PR #326. This makes sure we can keep `-Werror` in pipeline builds (which is beneficial for sdbus-c++ quality) and googletest always is pre-installed as a dependency, as opposed to being built within sdbus-c++ build.

Additionally, #326 also fixes the clang issue which we have now in this PR pipeline.

> IMHO there's no need for a squash since my PR contains three commits which do not need be squashed. IMHO squashing is not a good technique, because it get's rid of the commit history. Say for example your commit [9cc74e9](https://github.com/Kistler-Group/sdbus-cpp/commit/9cc74e9942a531a3db88d054a79631ecd8f8c67d) is so big, I can't really follow it and sometimes it lacks information on why you did certain things the way you did them.

I understand your sentiment, and agree. In many cases, however, it doesn't make sense, or it is impossible as such, to divide new functionality into several commits. It makes sense to bring it as one atomic commit, because all changes in that commit relate to each other. Take [9cc74e9](https://github.com/Kistler-Group/sdbus-cpp/commit/9cc74e9942a531a3db88d054a79631ecd8f8c67d), for example. Introduced sd-event support is a self-contained functionality, delivered as one piece. It is unnatural to me, and unnecessarily complicated, to break this into various commits. The functionality and its tests are delivered together. But, I understand your point about large commit. Hence, as a trade-off, this could have been potentially split into feature commit and tests commit. Such a division line would make sense to me.

> Happy to help every time. I appreciate the work you put into this library and really love using it.

Thanks for your contribution and help. And I'm glad to see people like to use the library; I did my best to design it for clients to enjoy using it.

hellow554:
#326 makes sense to me.
So, if you think you're ready, merge it and I'll drop my commit.

sangelovic:
@hellow554 done

PR #327, opened 2023-05-22
Prevent installing a time event, when there's no need for a timeout (cont.)

### LINKED ISSUE #324: New Integration of sd-event loop

hellow554:
I accidentally closed the other issue by wrongly deleting my branch... github is somewhat picky about that...

Fixes #324

Continues #325

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 
 cmake_minimum_required(VERSION 3.14)
 
-project(sdbus-c++ VERSION 1.2.0 LANGUAGES C CXX)
+project(sdbus-c++ VERSION 2.0.0 LANGUAGES C CXX)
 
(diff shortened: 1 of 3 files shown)


sangelovic:
Thanks

Issue #328, opened 2023-05-25
Use more std types


hellow554:
**Is your feature request related to a problem? Please describe.**

A lot of [dbus types](https://github.com/Kistler-Group/sdbus-cpp/blob/master/docs/using-sdbus-c++.md#using-d-bus-types) map to *std* types very nicely, e.g. all of the integer, float and string types.  
Then there are some special types, e.g. `OBJECT_PATH` and `SIGNATURE` where making a custom type makes a lot of sense, although there are simple wrappers around `std::string`.

1. Then there's ARRAY.
Currently only `std::vector` is accepted.<br>
What I want to see is, to also accept [`std::array`](https://en.cppreference.com/w/cpp/container/array) and [`std::span`](https://en.cppreference.com/w/cpp/container/span), because currently I have to convert those to an heap allocated vector, although I don't really need to.

2.  STRUCT
I mentioned it once in #164, that I want to see [`std::tuple`](https://en.cppreference.com/w/cpp/utility/tuple) as an accepted type.
The reason for that is the variadic arguments inside of `sdbus::Struct`.<br>
There are a lot of specialized functions for `std::tuple`, e.g. [`std::tuple_size`](https://en.cppreference.com/w/cpp/utility/tuple/tuple_size), [`std::tuple_element`](https://en.cppreference.com/w/cpp/utility/tuple/tuple_element), [`std::get`](https://en.cppreference.com/w/cpp/utility/tuple/get), [`std::make_tuple`](https://en.cppreference.com/w/cpp/utility/tuple/make_tuple) and [structured binding (!)](https://en.cppreference.com/w/cpp/language/structured_binding) probably many more I don't know.  
Those are **not** defined for `sdbus::Struct` and therefore a lot of template magic doesn't work for this type.  
You often find solutions on stack overflow that work for `std::tuple`, but you need some effort to make them work with `sdbus::Struct`.<br><br>
Therefore, I plead to deprecate/remove `sdbus::Struct` in `release/v2.0` and switch to `std::tuple`. I think it's a good time with a new major version to introduce new stuff.

Please let me know of what you think and how we can work together to make this work.

Thanks

hellow554:
ping @sangelovic 

Do you think it's worth to implement that? What are your thoughts? I'm more than willing to help you with this :)

sangelovic:
@hellow554 Ping received ;) I'll come back to you here today or tomorrow.

sangelovic:
@hellow554 

[...]

PR #342, opened 2023-08-02
feat: support serialization of array, span and unordered_map

sangelovic:
Add support for `std::array`, `std::span` and `std::unordered_map` in D-Bus message serialization/deserialization API.

TBD: Extending codegen XML to allow clients to choose from STL containers to use in C++ bindings?

### DIFF
--- a/include/sdbus-c++/Message.h
+++ b/include/sdbus-c++/Message.h
@@ -31,7 +31,12 @@
 #include <sdbus-c++/Error.h>
 #include <string>
 #include <vector>
+#include <array>
+#if __cplusplus >= 202002L
+#include <span>
+#endif
 #include <map>
+#include <unordered_map>
 #include <utility>
 #include <cstdint>
 #include <cassert>
@@ -85,10 +90,23 @@ namespace sdbus {
         Message& operator<<(const ObjectPath &item);
         Message& operator<<(const Signature &item);
         Message& operator<<(const UnixFd &item);
-        template <typename _Element> Message& operator<<(const std::vector<_Element>& items);
-        template <typename _Key, typename _Value> Message& operator<<(const std::map<_Key, _Value>& items);
-        template <typename... _ValueTypes> Message& operator<<(const Struct<_ValueTypes...>& item);
-        template <typename... _ValueTypes> Message& operator<<(const std::tuple<_ValueTypes...>& item);
+
+        template <typename _Element, typename _Allocator>
+        Message& operator<<(const std::vector<_Element, _Allocator>& items);
+        template <typename _Element, std::size_t _Size>
+        Message& operator<<(const std::array<_Element, _Size>& items);
+#if __cplusplus >= 202002L
+        template <typename _Element, std::size_t _Extent>
+        Message& operator<<(const std::span<_Element, _Extent>& items);
+#endif
+        template <typename _Key, typename _Value, typename _Compare, typename _Allocator>
+        Message& operator<<(const std::map<_Key, _Value, _Compare, _Allocator>& items);
+        template <typename _Key, typename _Value, typename _Hash, typename _KeyEqual, typename _Allocator>
+        Message& operator<<(const std::unordered_map<_Key, _Value, _Hash, _KeyEqual, _Allocator>& items);
+        template <typename... _ValueTypes>
+        Message& operator<<(const Struct<_ValueTypes...>& item);
+        template <typename... _ValueTypes>
+        Message& operator<<(const std::tuple<_ValueTypes...>& item);
 
         Message& operator>>(bool& item);
         Message& operator>>(int16_t& item);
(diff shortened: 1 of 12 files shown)


dleeds-cpi (commented):
I can't wait to make use of these additions!

include/sdbus-c++/Message.h
```diff
@@ -85,10 +90,23 @@ namespace sdbus {
...
+        Message& operator<<(const std::span<_Element, _Extent>& items);
+#endif
+        template <typename _Key, typename _Value, typename _Compare, typename _Allocator>
+        Message& operator<<(const std::map<_Key, _Value, _Compare, _Allocator>& items);
+        template <typename _Key, typename _Value, typename _Hash, typename _KeyEqual, typename _Allocator>
+        Message& operator<<(const std::unordered_map<_Key, _Value, _Hash, _KeyEqual, _Allocator>& items);
+        template <typename... _ValueTypes>
+        Message& operator<<(const Struct<_ValueTypes...>& item);
+        template <typename... _ValueTypes>
+        Message& operator<<(const std::tuple<_ValueTypes...>& item);
```
hellow554, line 109: So, you do allow std::tuple? I thought the agreement was to not allow that and use `sdbus::Struct` instead.

include/sdbus-c++/Message.h
```diff
@@ -105,10 +123,22 @@ namespace sdbus {
...
+        Message& operator>>(std::span<_Element, _Extent>& items);
+#endif
+        template <typename _Key, typename _Value, typename _Compare, typename _Allocator>
+        Message& operator>>(std::map<_Key, _Value, _Compare, _Allocator>& items);
+        template <typename _Key, typename _Value, typename _Hash, typename _KeyEqual, typename _Allocator>
+        Message& operator>>(std::unordered_map<_Key, _Value, _Hash, _KeyEqual, _Allocator>& items);
+        template <typename... _ValueTypes>
+        Message& operator>>(Struct<_ValueTypes...>& item);
+        template <typename... _ValueTypes>
+        Message& operator>>(std::tuple<_ValueTypes...>& item);
```
hellow554, line 141: Same here

hellow554:
At at first glance, this looks very good to me :) Thanks for that! (Better than I would have done, so I'm glad you did it :D )

Just the `std::tuple` thing bugs me a little bit.

Also I would like to see the documentation (https://github.com/Kistler-Group/sdbus-cpp/blob/master/docs/using-sdbus-c%2B%2B.md) updated, but I can do that if you like

include/sdbus-c++/Message.h
sangelovic: Yes, what you say holds. We use `Struct`. Nothing changed here regarding `operator<<(const std::tuple&)`, I just broke the declaration in two lines, hence it appears as a new stuff. The tuple operator there and needs to stay there, too, because it's used and needed by sdbus-c++ and its clients. It serves different purpose than `Struct`. See the a) and b) points in a comment of mine where I explained that: https://github.com/Kistler-Group/sdbus-cpp/issues/328#issuecomment-1651602916

include/sdbus-c++/Message.h
sangelovic: Same reply above

sangelovic:
@hellow554 Thanks for your support here. Documentation is a good point, I forgot about it ;-) It's probably not much to update there? Would you contribute here? Will we create a new follow PR? How much time do you need?

hellow554:
https://gist.github.com/hellow554/2a23cb29b87e4c6cac7f58d7aa96ed55

Please see this gist and apply them after you downloaded them with `git am 0001-fix-some-spelling-mistakes.patch && git am 0002-docs-update-table-of-valid-c-types.patch`

Thanks a lot :)

sangelovic:
Done, thanks!

sangelovic:
@hellow554

What about your open comments? Still something to discuss, or can we resolve?

And there is still one point open -- shall we also extend the code generator to be able to generate these additional containers depending on the client's choice in the D-Bus XML? I guess it will make the change consistent across all levels of API and make code generator more flexible and usable in those cases.

hellow554:
> What about your open comments? Still something to discuss, or can we resolve?

Everthing is fine :)

> shall we also extend the code generator to be able to generate these additional containers depending on the client's choice in the D-Bus XML

TBH: I never used the XML generator, although I probably should have :D
I can't tell for sure, but it sounds like a good idead.
Maybe something like this:

`<arg type="ai" container="std::array" name="numbers" direction="in" />`

But I'm not sure if this is a feasible things, because how does the code generator know the size of the array? Do you have to tell them, e.g. `container="std::array<5>"`, but that feels weird, because normally you specify the type and then the length, but the type is deduced by the `i` after the `a`, isn't it?

TBH: I'm not sure how to do that in the XML and it needs some dicussion, but I don't think that it should be disucess in this PR.

I would open a new issue where one/we/other can discuss the layout of the XML.

sangelovic:
@hellow554

We could probably use another optional XML attribute named e.g. `size` which would denote container size: `<arg type="ai" container="std::array" size="5" name="numbers" direction="in" />`. Not sure however whether adding custom attributes is allowed or whether it breaks validators or such things. Another way is to use annotations (these can be extended). For example:

```xml
<arg type="ai" name="numbers" direction="in">
    <annotation name="org.freedesktop.SdBusCpp.ContainerType" value="std::array" />
    <annotation name="org.freedesktop.SdBusCpp.ContainerSize" value="5" />
</arg>
```

Either way, how to address complex types? E.g. `std::map<int, std::vector<std::map<std::string, std::vector<int>>>` is a type of a function argument. Shall all `std::vector`s be replaced with `std::array`s? This may be very limiting.

This needs further thinking, and I agree with you we shall do that into a separate issue/PR. We still may decide not to do that at all, if it would make things very complex.

Anyway, thanks for your contributions!

PR #343, opened 2023-08-02
feat: make Struct tuple-like class

sangelovic:
This has `Struct` implement tuple protocol, making `Struct` usable virtually anywhere tuple is, i.e. `std::get<>()`, `std::tuple_element`, `std::tuple_size`, and in structured bindings.


include/sdbus-c++/Types.h
```diff
@@ -280,4 +287,14 @@ namespace sdbus {
 
 }
 
+template <size_t _I, typename... _ValueTypes>
+struct std::tuple_element<_I, sdbus::Struct<_ValueTypes...>>
+        : std::tuple_element<_I, std::tuple<_ValueTypes...>>
+{};
```
hellow554, line 293: This is an amazing neat solution. I would have never come up with this solution. Amazing :)

hellow554:
LGTM.
As said, I try to test it with our current code and report today or tomorrow back if it works. Give me 24 hours ;)

include/sdbus-c++/Types.h
sangelovic: Yes, neat, nice, simple (and standard-conforming) solution to make any class a tuple-like class :)

sangelovic:
@hellow554 OK

hellow554:
I was able to test your code and it works flawlessly.
I can remove my previous workarounds and everything is fine :)

Thanks a lot for this!

IMHO this is ready to be merged

PR #350, opened 2023-08-16
feat: add support for direct connections

fedyarov:
Hi there!

This commit adds support for create direct connections with sdbus-cpp.

There is a design issue over which i though for a while. In other libraries for DBus such as `libdbus`, `QtDBus`, `GDBus` exists entity `DBusServer` that provide functionality for creating direct connections. User specify only address with desired transport (https://dbus.freedesktop.org/doc/dbus-specification.html#Transports). But in `sdbus` there is no such convenient entity and user have to create listening server by his own. He needs to create bus with `sd_bus_new` and then configure that bus with  `sd_bus_set_fd` call, when `fd` is the file descriptor for separate connection (returned by the  `accept` syscall, for instance). Examples can be found here https://www.spinics.net/lists/systemd-devel/msg01558.html. I decided to follow their path and leave the responsibility of creating a listening socket to user.

I tried to found a way to provide convenient `DBusServer` class. But such class requires an general event loop in which to be integrated. Event loop in `Connection` class not suitable, because it polling it's own `sd_bus` bus.

There also a way to create api method `std::unique_ptr<sdbus::IConnection> createServerBus(const std::string &address)` where user can pass address with desired transport. But this method must block until client connect, because in other way `BusPtr` inside `Connection` will be invalid until client connected and `sd_bus_start` called.

### DIFF
--- a/docs/using-sdbus-c++.md
+++ b/docs/using-sdbus-c++.md
@@ -21,7 +21,8 @@ Using sdbus-c++ library
 16. [Standard D-Bus interfaces](#standard-d-bus-interfaces)
 17. [Using D-Bus Types](#using-d-bus-types)
 18. [Support for match rules](#support-for-match-rules)
-19. [Conclusion](#conclusion)
+19. [Using direct (peer-to-peer) D-Bus connections](#using-direct-peer-to-peer-d-bus-connections)
+20. [Conclusion](#conclusion)
 
 Introduction
 ------------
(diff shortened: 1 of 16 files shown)


fedyarov:
Hi @sangelovic !
Can you please review this and leave your comments?

sangelovic:
Hello Maksim, thanks for the PR, yes I will review, hopefully soon. I worked on other priorities and now am out for a while 😉 Can't tell exactly when I find time.

-------- Original Message --------

sangelovic:
Hi @fedyarov! I know it took longer, but I've finally made it to your PR :o) I went through and I did some modifications, most notably:

* I added `createDirectBusConnection()` overload with `int fd`, in addition to `const std::string& address`.
* I changed the comment about ownership of server `fd`. In fact, the library takes over `fd` ownership, and when the connection instance is destroyed, the `fd` gets closed automatically. The same behavior applies to client connection. Whole library is built around principle of RAII and resource ownership.
* Instead of `std::optional<std::string>` type for destination parameter, I changed back to plain `std::string`. I think it's an overkill to change that type just because of a narrow minority of cases of direct bus connections. Second, it breaks ABI compatibility. We can easily emulate the condition of empty destination parameter with a sentinel value of empty string. I've done it that way.
* I refactored the integration test. There is a new fixture into which I factored out all the helper code for preparing the connections. And the core of test is now the verification of a simple D-Bus function call and a D-Bus signal emission. I think that the communication works is essential to verify.

Please feel free to view my changes and let me know what do you think. If you agree, I can merge it right away.

Thank you for your contribution, and also for your patience!

dreiss:
One thing I'm noticing with this feature (as well as the version I submitted in #355) is that it causes a crash if the peer goes away:

```
  what():  [org.freedesktop.DBus.Error.Disconnected] Failed to process bus requests (Connection reset by peer)
*** Aborted at 1696890980 (Unix time, try 'date -d @1696890980') ***
*** Signal 6 (SIGABRT) (0x5e126) received by PID 385318 (pthread TID 0x7f0842200640) (linux TID 385502) (maybe from PID 385318, UID 0) (code: -6), stack trace: ***
    @ 000000000005593c sdbus::internal::Connection::processPendingRequest()
                       ./third-party/sdbus-cpp/src/Connection.cpp:500
    @ 0000000000054b66 sdbus::internal::Connection::enterEventLoop()
                       ./third-party/sdbus-cpp/src/Connection.cpp:136
```

It would be great to have a way to configure the connection to handle this more gracefully.  In my case, the client is a unit test and the server is a subprocess, so my test crashes during shutdown after I clean up the server.  I'm working around this by calling `getConnection().leaveEventLoop()`, but it's a bit invasive.

sangelovic:
@dreiss

Yes, I'm aware of this behavior. The integration tests for direct connections do stop the event loops explicitly, too, before destroying connection objects. The questions is whether to try to hide this from users or whether to make it a documented behavior that clients using direct connections need to be aware of. Hiding it is is not so easy, AFAICS. The connection would either need to know its "type" -- that it is a direct connection. The beauty of current design is that individual types of connections do differ only in their construction strategy (through c-tors with tags), otherwise their behavior is all the same. 

Additionally, even if the connection knew it is a direct connection, there is a way to create a connection from an `sd_bus` instance. And here we have no idea what kind of `sd_bus` (whether direct or D-Bus daemon-based one) instance is passed, which renders this solution impossible.

Another approach would be kind to `Connection reset by peer` errors and instead of throwing an exception, ignore the error, process nothing, and go on polling for events. But this is a relaxation that would hold of all types of connections and I think this sounds like over-relaxing.

Do you have more ideas?

For the moment, I think it's an acceptable trade-off solution that such a destruction workflow is required and documented (it's anyway a special, rare case of direct connections). I've updated the documentation in aac7e590ea497ff05bf55e1f68713df77d9e5905.

PR #352, opened 2023-08-18
build: Add support for elogind.

Apteryks:
* CMakeLists.txt: Fallback to elogind when libsystemd could not be found.


Apteryks:
@sangelovic thank you!

sangelovic:
@Apteryks 👍 thank you for contribution

PR #355, opened 2023-08-26
feat: Add support for peer-to-peer client connections

dreiss:
This is for programs that want to use the D-Bus protocol for their connections, but don't have a D-Bus daemon, ObjectManager, etc.

See https://dbus2.github.io/zbus/connection.html#peer-to-peer-connection for zbus's (brief) documentation of this technique.  sd_bus covers it at https://www.freedesktop.org/software/systemd/man/sd_bus_set_server.html#

### DIFF
--- a/include/sdbus-c++/IConnection.h
+++ b/include/sdbus-c++/IConnection.h
@@ -458,6 +458,18 @@ namespace sdbus {
      * @throws sdbus::Error in case of failure
      */
     [[nodiscard]] std::unique_ptr<sdbus::IConnection> createRemoteSystemBusConnection(const std::string& host);
+
+    /*!
+     * @brief Creates/opens a peer-to-peer D-Bus connection at a custom address
+     *
+     * @param[in] address ";"-separated list of addresses to try to connect
+     * @return Connection instance
+     *
+     * @throws sdbus::Error in case of failure
+     *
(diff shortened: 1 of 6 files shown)


dreiss:
I'd like to get some feedback on this feature, as I'm testing it for a project within Meta.

How would you like to see this tested?  For my project, I'm using zbus as the server, but I could do a little server for tests using the raw sd_bus API.

sangelovic:
Hi @dreiss! Exactly that functionality was proposed in #350 PR shortly before yours. It includes more -- support for creation of a D-Bus server, and an integration test case for such a direct connection and communication (which I think also answers your question about testing it).

Support for creating a D-Bus server is a very simple one. There are a few other tweaks and settings that raw sd-bus API provides when creating a server connection. We don't. In order to avoid having to provide C++ convenience wrappers for all these sd-bus API functions, I plan to introduce a `Connection` constructor that simply consumes the client-provided `sd_bus` instance and assumes its ownership. This way, clients will be able to create `sd_bus` and tweak it as they wish, and then pass it over to `Connection` and then work with `Connection` only.

If you agree I will close this PR as a duplicate of #350. If you'd like to review #350 and leave comments, feel free to do so. Thank you!

sangelovic:
Duplicate to #350

sangelovic:
Duplicate to #350

PR #358, opened 2023-09-08
Add FreeBSD support

jbeich:
Tested [downstream](https://www.freshports.org/devel/sdbus-cpp) via [pass-secrets](https://github.com/nullobsi/pass-secrets) and [xdg-desktop-portal-hyprland](https://github.com/hyprwm/xdg-desktop-portal-hyprland).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,3 +99,7 @@ jobs:
         run: |
           cmake -B _build -G Ninja -DBUILD_CODE_GEN=ON -DBUILD_TESTS=ON -DENABLE_PERF_TESTS=ON -DENABLE_STRESS_TESTS=ON
           cmake --build _build
+          cmake --install _build
+          pkg install -y dbus
+          service dbus onestart
+          ctest --output-on-failure --test-dir _build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,19 +48,19 @@ jobs:
       run: |
         mkdir build
         cd build
-        cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS="-O0 -g -W -Wextra -Wall -Wnon-virtual-dtor -Werror" -DCMAKE_VERBOSE_MAKEFILE=ON -DBUILD_TESTS=ON -DENABLE_PERF_TESTS=ON -DENABLE_STRESS_TESTS=ON -DBUILD_CODE_GEN=ON ..
+        cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX:PATH=/usr -DCMAKE_CXX_FLAGS="-O0 -g -W -Wextra -Wall -Wnon-virtual-dtor -Werror" -DCMAKE_VERBOSE_MAKEFILE=ON -DBUILD_TESTS=ON -DENABLE_PERF_TESTS=ON -DENABLE_STRESS_TESTS=ON -DBUILD_CODE_GEN=ON ..
     - name: configure-release
       if: matrix.build == 'shared-libsystemd' && matrix.os == 'ubuntu-22.04'
       run: |
         mkdir build
         cd build
-        cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_FLAGS="-O3 -DNDEBUG -W -Wextra -Wall -Wnon-virtual-dtor -Werror" -DCMAKE_VERBOSE_MAKEFILE=ON -DBUILD_TESTS=ON -DENABLE_PERF_TESTS=ON -DENABLE_STRESS_TESTS=ON -DBUILD_CODE_GEN=ON ..
+        cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX:PATH=/usr -DCMAKE_CXX_FLAGS="-O3 -DNDEBUG -W -Wextra -Wall -Wnon-virtual-dtor -Werror" -DCMAKE_VERBOSE_MAKEFILE=ON -DBUILD_TESTS=ON -DENABLE_PERF_TESTS=ON -DENABLE_STRESS_TESTS=ON -DBUILD_CODE_GEN=ON ..
     - name: configure-with-embedded-libsystemd
       if: matrix.build == 'embedded-static-libsystemd'
       run: |
         mkdir build
         cd build
-        cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE=ON -DBUILD_TESTS=ON -DENABLE_PERF_TESTS=ON -DENABLE_STRESS_TESTS=ON -DBUILD_CODE_GEN=ON -DBUILD_LIBSYSTEMD=ON -DLIBSYSTEMD_VERSION=244 ..
+        cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX:PATH=/usr -DCMAKE_VERBOSE_MAKEFILE=ON -DBUILD_TESTS=ON -DENABLE_PERF_TESTS=ON -DENABLE_STRESS_TESTS=ON -DBUILD_CODE_GEN=ON -DBUILD_LIBSYSTEMD=ON -DLIBSYSTEMD_VERSION=244 ..
     - name: make
       run: |
         cd build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,3 +84,18 @@ jobs:
           build/sdbus-c++*.deb
           build/sdbus-c++*.ddeb
         retention-days: 10
+  freebsd-build:
+    name: build (freebsd, clang/libc++, basu)
+    runs-on: macos-12 # until https://github.com/actions/runner/issues/385
+    steps:
+    - uses: actions/checkout@v2
+    - name: Test in FreeBSD VM
+      uses: vmactions/freebsd-vm@v0
+      with:
+        copyback: false
+        usesh: true
+        prepare: |
+          pkg install -y cmake ninja pkgconf basu expat googletest
+        run: |
+          cmake -B _build -G Ninja -DBUILD_CODE_GEN=ON -DBUILD_TESTS=ON -DENABLE_PERF_TESTS=ON -DENABLE_STRESS_TESTS=ON
+          cmake --build _build
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,12 @@ if(NOT BUILD_LIBSYSTEMD)
             set(LIBSYSTEMD "elogind")
             string(REPLACE "." ";" VERSION_LIST ${Systemd_VERSION})
             list(GET VERSION_LIST 0 Systemd_VERSION)
+	else()
+            message(WARNING "libelogind not found, checking for basu instead")
+            pkg_check_modules(Systemd IMPORTED_TARGET GLOBAL basu)
+            set(LIBSYSTEMD "basu")
+            # https://git.sr.ht/~emersion/basu/commit/d4d185d29a26
+            set(Systemd_VERSION "240")
         endif()
     endif()
     if(NOT TARGET PkgConfig::Systemd)
@@ -119,6 +125,7 @@ add_library(sdbus-c++-objlib OBJECT ${SDBUSCPP_SRCS})
 target_compile_definitions(sdbus-c++-objlib PRIVATE
     BUILD_LIB=1
     LIBSYSTEMD_VERSION=${LIBSYSTEMD_VERSION}
+    SDBUS_${LIBSYSTEMD}
     SDBUS_HEADER=<${LIBSYSTEMD}/sd-bus.h>)
 target_include_directories(sdbus-c++-objlib PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
                                                    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ include(GNUInstallDirs) # Installation directories for `install` command and pkg
 # PERFORMING CHECKS & PREPARING THE DEPENDENCIES
 #-------------------------------
 
-set(LIBSYSTEMD "libsystemd")
+set(LIBSYSTEMD "systemd")
 
 option(BUILD_LIBSYSTEMD "Build libsystemd static library and incorporate it into libsdbus-c++" OFF)
 
@@ -23,7 +23,7 @@ if(NOT BUILD_LIBSYSTEMD)
         message(WARNING "libsystemd not found, checking for libelogind instead")
         pkg_check_modules(Systemd IMPORTED_TARGET GLOBAL libelogind>=236)
         if(TARGET PkgConfig::Systemd)
-            set(LIBSYSTEMD "libelogind")
+            set(LIBSYSTEMD "elogind")
             string(REPLACE "." ";" VERSION_LIST ${Systemd_VERSION})
             list(GET VERSION_LIST 0 Systemd_VERSION)
         endif()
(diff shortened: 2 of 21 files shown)


CMakeLists.txt
```diff
@@ -23,9 +23,14 @@ if(NOT BUILD_LIBSYSTEMD)
         message(WARNING "libsystemd not found, checking for libelogind instead")
         pkg_check_modules(Systemd IMPORTED_TARGET GLOBAL libelogind>=236)
         if(TARGET PkgConfig::Systemd)
-            set(LIBSYSTEMD "libelogind")
+            set(LIBSYSTEMD "elogind")
             string(REPLACE "." ";" VERSION_LIST ${Systemd_VERSION})
             list(GET VERSION_LIST 0 Systemd_VERSION)
+	else()
+            pkg_check_modules(Systemd IMPORTED_TARGET GLOBAL basu)
```
sangelovic, line 30: I suggest adding a short comment here about `basu` (what it is) and FreeBSD support, and also add `message(WARNING "libelogind not found, checking for basu instead")` like above in the file

CMakeLists.txt
```diff
@@ -116,7 +121,11 @@ endif()
 
 # Having an object target allows unit tests to reuse already built sources without re-building
 add_library(sdbus-c++-objlib OBJECT ${SDBUSCPP_SRCS})
-target_compile_definitions(sdbus-c++-objlib PRIVATE BUILD_LIB=1 LIBSYSTEMD_VERSION=${LIBSYSTEMD_VERSION})
+target_compile_definitions(sdbus-c++-objlib PRIVATE
+    BUILD_LIB=1
+    LIBSYSTEMD_VERSION=${LIBSYSTEMD_VERSION}
+    SDBUS_${LIBSYSTEMD}
+    SDBUS_HEADER=<${LIBSYSTEMD}/sd-bus.h>)
```
sangelovic, line 128: I propose making sd-bus include folder variable, so instead of `SDBUS_HEADER` we'd have `SDBUS_INCLUDE_DIR`:

`SDBUS_INCLUDE_DIR=${LIBSYSTEMD}`

and then the usage in remaining files would be:

`#include <SDBUS_INCLUDE_PATH/sd-bus.h>`

tests/CMakeLists.txt
```diff
@@ -104,11 +104,15 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 #----------------------------------
 
 add_executable(sdbus-c++-unit-tests ${UNITTESTS_SRCS})
-target_compile_definitions(sdbus-c++-unit-tests PRIVATE LIBSYSTEMD_VERSION=${LIBSYSTEMD_VERSION})
+target_compile_definitions(sdbus-c++-unit-tests PRIVATE
+    LIBSYSTEMD_VERSION=${LIBSYSTEMD_VERSION}
+    SDBUS_HEADER=<${LIBSYSTEMD}/sd-bus.h>)
```
sangelovic, line 109: Do we need this? I don't see sd-bus headers to be used anywhere in the unit and integration tests...

tests/CMakeLists.txt
```diff
@@ -104,11 +104,15 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR})
...
 target_link_libraries(sdbus-c++-unit-tests sdbus-c++-objlib GTest::gmock)
 
 add_executable(sdbus-c++-integration-tests ${INTEGRATIONTESTS_SRCS})
-target_compile_definitions(sdbus-c++-integration-tests PRIVATE LIBSYSTEMD_VERSION=${LIBSYSTEMD_VERSION})
+target_compile_definitions(sdbus-c++-integration-tests PRIVATE
+    LIBSYSTEMD_VERSION=${LIBSYSTEMD_VERSION}
+    SDBUS_HEADER=<${LIBSYSTEMD}/sd-bus.h>)
```
sangelovic, line 115: Do we need this? I don't see sd-bus headers to be used anywhere in the unit and integration tests...

sangelovic (changes requested):
Thanks for the contribution! Glad to enrich also the FreeBSD world with this lib. I made some comments and asked for changes. Will you address them, or shall I?

tests/CMakeLists.txt
jbeich: It's used indirectly:
```c++
In file included from tests/unittests/Connection_test.cpp:28:
In file included from src/Connection.h:32:
src/IConnection.h:31:10: error: expected "FILENAME" or <FILENAME>
#include SDBUS_HEADER
         ^
src/IConnection.h:61:59: error: unknown type name 'sd_bus_vtable'
                                                  , const sd_bus_vtable* vtable
                                                          ^
[...]
```

.github/workflows/ci.yml
```diff
@@ -84,3 +84,18 @@ jobs:
...
+    - name: Test in FreeBSD VM
+      uses: vmactions/freebsd-vm@v0
+      with:
+        copyback: false
+        usesh: true
+        prepare: |
+          pkg install -y cmake ninja pkgconf basu expat googletest
+        run: |
+          cmake -B _build -G Ninja -DBUILD_CODE_GEN=ON -DBUILD_TESTS=ON -DENABLE_PERF_TESTS=ON -DENABLE_STRESS_TESTS=ON
+          cmake --build _build
```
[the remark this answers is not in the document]
jbeich: - 1 integration test hangs since 1.3.0 (haven't filed a bug yet)
-  `etc/dbus-1/system.d/org.sdbuscpp.integrationtests.conf` requires system bus (dbus daemon running as root)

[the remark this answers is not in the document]
jbeich: Also, I didn't fix installation of tests yet: `/etc/dbus-1` doesn't respect `CMAKE_INSTALL_PREFIX` and `/opt/test/bin` doesn't respect hier(7).

CMakeLists.txt
jbeich: - basu doesn't provide any other header except `sd-bus.h`
- `<SDBUS_INCLUDE_PATH/sd-bus.h>` isn't accepted by C preprocessor

```c
$ echo '#include <DIR/types.h>' >a.c
$ cc -DDIR=sys -c a.c
a.c:1:10: fatal error: 'DIR/types.h' file not found
#include <DIR/types.h>
         ^~~~~~~~~~~~~
1 error generated.
```

jbeich: > basu doesn't provide any other header except `sd-bus.h`

Nevermind, I misremembered:
```
$ pkg info -l basu
basu-0.2.1:
        /usr/local/bin/basuctl
        /usr/local/include/basu/_sd-common.h
        /usr/local/include/basu/sd-bus-protocol.h
        /usr/local/include/basu/sd-bus-vtable.h
        /usr/local/include/basu/sd-bus.h
        /usr/local/include/basu/sd-id128.h
        /usr/local/lib/libbasu.so
        /usr/local/lib/libbasu.so.0
        /usr/local/libdata/pkgconfig/basu.pc
        /usr/local/share/licenses/basu-0.2.1/LGPL21+
        /usr/local/share/licenses/basu-0.2.1/LICENSE
        /usr/local/share/licenses/basu-0.2.1/catalog.mk

```

tests/CMakeLists.txt
jbeich: No (or not yet), so reverted and squashed.

CMakeLists.txt
jbeich: - explaining `basu` wouldn't be consistent with `elogind`
- added (squashed) `message(...)` as suggested

jbeich:
Beware, [`release/v2.0`](https://github.com/Kistler-Group/sdbus-cpp/tree/release/v2.0) contains 9cc74e9942a5 which isn't supported by either `basu` or ConsoleKit2 (sd-compat.c). `*SdEvent` API will have to be disabled on FreeBSD via `#ifndef SDBUS_basu` or similar.

CMakeLists.txt
jbeich: FWIW, `basu` support isn't FreeBSD-specific. Many Linux distros have it packaged, see https://repology.org/project/basu/versions

.github/workflows/ci.yml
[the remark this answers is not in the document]
sangelovic: To the respective points:

1. May be an issue beyond sdbus-c++ (see https://github.com/Kistler-Group/sdbus-cpp/issues/359#issuecomment-1721959843). Or will be fixed with sdbus-c++ v2.0 (once `release/v2.0` is merged into `master`).
2. Is this really an issue? D-Bus daemon with system bus on my native system as well as embedded system runs under `messagebus` user. This is configured in `/usr/share/dbus-1/system.conf` (as per `man dbus-daemon`). Alternatively, integration tests could be modified to run on session bus. But not all systems (like the embedded one of mine) run D-Bus daemon with session bus.
3. `/opt/test/bin` is a default value of `TESTS_INSTALL_PATH` variable, and thus can be overridden. However, true, `/etc/dbus-1` is fully hardcoded absolute path. We could fix this in this PR, by removing the `/` from the beginning of those paths.

tests/CMakeLists.txt
sangelovic: OK, we only need this for unit tests.

CMakeLists.txt
sangelovic: OK, let's keep it as it is.

.github/workflows/ci.yml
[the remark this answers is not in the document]
jbeich: > 1\. May be an issue beyond sdbus-c++

Or an undefined behavior in sdbus-cpp exposed by a different threading library. I'm mainly a packager, not familar with debugging threads.

> 2\. Is this really an issue?

Somewhat:
- FreeBSD doesn't ship with DBus in base system. Downstream (packaging) only allows root during installation of dependencies.
- Many DBus consumers (both general and some sdbus-cpp) only need user/session bus thus work fine without extra privileges required to run system bus.
- Services on system bus are more destructive thus not a good idea to run multiple instances or pass through socket from host to jail.
- Installing system bus configs under `etc/dbus-1/system.d` requires `dbus-daemon` if the directory didn't exist before.
- System bus isn't required on non-systemd distros: non-existent on servers, unused besides bluetooth (bluez), battery status (upower) and session tracking (elogind, consolekit2) on desktop.

Looks like `freebsd-vm` runs everything as root, so I've enabled the tests.

> 3\. `/opt/test/bin` is a default value of `TESTS_INSTALL_PATH` variable, and thus can be overridden.

This shouldn't be required by default. Fixed in d81ab4ba30a2

> We could fix this in this PR, by removing the `/` from the beginning of those paths.

This breaks Linux CI (`org.sdbuscpp.integrationtests.conf` is now unused), so fixed by b748634d1de2 + 815190361ed4.

tests/CMakeLists.txt
```diff
@@ -138,21 +140,22 @@ endif()
 # INSTALLATION
 #----------------------------------
 
-set(TESTS_INSTALL_PATH "/opt/test/bin" CACHE STRING "Specifies where the test binaries will be installed")
+include(GNUInstallDirs)
+set(TESTS_INSTALL_PATH "tests" CACHE STRING "Specifies where the test binaries will be installed")
```
sangelovic, line 144: Is `tests` dir a kind of standard in packaging? When installing into `/`, it would be `/tests`, when installing into `/usr`, it would be `/usr/tests`.

tests/CMakeLists.txt
```diff
@@ -138,21 +140,22 @@ endif()
...
 
 install(TARGETS sdbus-c++-unit-tests DESTINATION ${TESTS_INSTALL_PATH} COMPONENT test)
 install(TARGETS sdbus-c++-integration-tests DESTINATION ${TESTS_INSTALL_PATH} COMPONENT test)
-install(FILES ${INTEGRATIONTESTS_SOURCE_DIR}/files/org.sdbuscpp.integrationtests.conf DESTINATION /etc/dbus-1/system.d COMPONENT test)
+install(FILES ${INTEGRATIONTESTS_SOURCE_DIR}/files/org.sdbuscpp.integrationtests.conf DESTINATION ${CMAKE_INSTALL_FULL_SYSCONFDIR}/dbus-1/system.d COMPONENT test)
```
sangelovic, line 148: Just to make sure: You mentioned previously that `/etc/dbus-1` does not respect `CMAKE_INSTALL_PREFIX`, but this change doesn't respect it either. But I guess we don't want to respect it here, right?

tests/CMakeLists.txt
sangelovic: As an additional note, changing the default install path may break already existing packagers. But I don't want to create a new major release just for that. I think such a change could be acceptable with migration from `v1.3` to `v1.4`. What do you think from your packager POV?

tests/CMakeLists.txt
jbeich: It does and confirmed by CI: integration tests on Ubuntu and FreeBSD would fail otherwise. `${CMAKE_INSTALL_FULL_SYSCONFDIR}` expands to `${CMAKE_INSTALL_PREFIX}/etc` or `/etc` if `${CMAKE_INSTALL_PREFIX}` is `/usr` (common on Linux).

jbeich: To see CI results enable workflow on this PR or for all PRs. I'm not gonna link to my results as I've removed `master` branch in my fork and plan to remove the whole fork once this is merged.

tests/CMakeLists.txt
jbeich: `/usr/tests/<project>` is standard (per [hier(7)](https://man.netbsd.org/hier.7) and [tests(7)](https://man.netbsd.org/tests.7)) on systems using Kyua like NetBSD and FreeBSD. Otherwise, tests are not usually installed. Some projects install in random places like `/usr/share/<project>/tests` (see `CMAKE_INSTALL_DATADIR`) or `/usr/share/examples/<project>`.

`BUILD_TESTS` is disabled by default, so packagers are unlikely to build, let alone, install tests. Looking at [Repology](https://repology.org/project/sdbus-cpp/versions) only Debian touches it but [skips](https://salsa.debian.org/zhsj/sdbus-cpp/-/blob/master/debian/not-installed) installing `opt/test/bin` in favor of using [dedicated system bus instance](https://salsa.debian.org/zhsj/sdbus-cpp/-/blob/bfd17ed50a31451cfd19e01fca544756e5c1b397/debian/rules#L19-29).

Prebuilt tests are not part of public API, so changing the install location seems fine for a minor version update. Even moving tests installation from `BUILD_TESTS` to `INSTALL_TESTS` should be fine.

jbeich: Completely forgot, some projects install tests under regular `/usr/bin` (aka `${CMAKE_INSTALL_BINDIR}`) but gate based on off-by-default option like `INSTALL_TESTS`.

jbeich: I've dropped `/opt/test/bin` change to avoid scope creep. The priority here is to get FreeBSD support rather than simplify packaging.

sangelovic: OK. Could you please submit a separate PR with proposed packaging simplifications? Including splitting into `BUILD_TESTS` and `INSTALL_TESTS` (and whatever else you are proposing to improve or simplify packaging of the library). Your contribution would be appreciated 👍 🙂

sangelovic:
@jbeich Do you want that I tag a new release? I would like to first finish open PRs about direct connections, unless you hurry

jbeich:
Waiting for a minor release is fine with me. Downstream [can backport](https://github.com/freebsd/freebsd-ports/commit/48cb5e90d88c) and uses system package in consumers.

sangelovic:
@jbeich I realized I'll have to rename the commit 🫤 because I forgot to adjust the name according to the conventional changelog when merging. But I don't know how to do that via GitHub web interface, and I'm out without my PC until tomorrow 🫢

sangelovic:
@jbeich So you will probably have to change the commit hash

jbeich:
GitHub supports referencing disconnected commits (don't belong to any branch), so URLs shouldn't change. Downstream it's more important to have the same patch content (instead of commit hash) and the reference to upstream review.

Now, I'm curious how are you going to to change anything already on the default branch (`master` here, `main` in new repos) without force-pushing or deleting and re-creating the branch.

jbeich:
Nevermind, I didn't notice the revert commit.

sangelovic:
@jbeich I fixed HEAD by force-pushing, please use the new hash: 1e2d13a04a988afac00929d837354bd4bcc2a450. Thanks and sorry for extra effort. This is also the reason why I asked for waiting for the new tag before referencing it downstream... (I could not dare to overwrite history of tagged version)

sangelovic:
@jbeich New release [v1.4.0](https://github.com/Kistler-Group/sdbus-cpp/releases/tag/v1.4.0) has just been published.

PR #361, opened 2023-09-18
fix: honor CMAKE_POSITION_INDEPENDENT_CODE when building

### LINKED ISSUE #357: Building statically but linking to a shared library results in linker error
**Describe the bug**
When compiling the library with `BUILD_SHARED_LIBS` set to `OFF`, and then linking to an output target that compiles their libraries as shared, we get linker errors due to the `-fPIC` option not being set in compile arguments.

**To Reproduce**
Create a CMake project that pulls in sdbus-cpp as a subproject, and compile that as static by setting `BUILD_SHARED_LIBS` to off. Then create a shared library in the top CMake project and link sdbus-cpp to it.

**Expected behavior**
In my project, I set `CMAKE_POSITION_INDEPENDENT_CODE` to `ON`. I expect this value to carry through to this project.

**Additional context**
OS: Arch Linux
SDBus version: v1.3.0

This line seems to be the culprit, setting the `POSITION_INDEPENDENT_CODE` variable to the value of `BUILD_SHARED_LIBS`
In this context, it is being set to off, even though this option is needed to link the statically compiled sdbus-c++ to shared libraries.
https://github.com/Kistler-Group/sdbus-cpp/blob/0eda85574546d19d9f06d6d5418bc192b3846f96/CMakeLists.txt#L123

sangelovic:
Fixes #357 

/cc @proatgam


russelltg:
I think it's sufficient to just do 
```
if(BUILD_SHARED_LIBS)
    set_target_properties(sdbus-c++-objlib PROPERTIES POSITION_INDEPENDENT_CODE ON)
enidf()
```

as POSITION_INDEPENDENT_CODE is automatically initialized by CMAKE_POSITION_INDEPENDENT_CODE

(source: https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html)

sangelovic:
> I think it's sufficient to just do
> 
> ```
> if(BUILD_SHARED_LIBS)
>     set_target_properties(sdbus-c++-objlib PROPERTIES POSITION_INDEPENDENT_CODE ON)
> enidf()
> ```
> 
> as POSITION_INDEPENDENT_CODE is automatically initialized by CMAKE_POSITION_INDEPENDENT_CODE
> 
> (source: https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html)

You are right. I simplified it. Thanks!

PR #362, opened 2023-09-23
fix: Use-after-return in synchronous calls

dreiss:
This bug was introduced by c39bc637b8df7791 and can be reproduced by configuring with

  cmake -S . -B build -DBUILD_TESTS=yes -DCMAKE_CXX_COMPILER=clang++ \
    -DCMAKE_CXX_FLAGS="-Wno-error=deprecated-copy -fno-omit-frame-pointer -fsanitize=address -fsanitize-address-use-after-scope"

and running `cmake --buid build && cmake --build build -t test` or `build/tests/sdbus-c++-integration-tests --gtest_filter=SdbusTestObject.HandlesCorrectlyABulkOfParallelServerSideAsyncMethods`

The issue is that `sdbus_async_reply_handler` can call `removeCall`, which writes to `data->finished`, but `data` can point to the stack of `sendMethodCallMessageAndWaitForReply`, which can return as soon as `asyncCallData->callback` is called.

As a fix, I restored some of the logic removed in c39bc637b8df7791. Specifically, in `sdbus_async_reply_handler`, I make a copy of some data from `asyncCallData` (a new `state` field instead of `slot`), and in the `SCOPE_GUARD`, I don't call `removeCall` if the call was actually synchronous.

### DIFF
--- a/src/Proxy.cpp
+++ b/src/Proxy.cpp
@@ -120,7 +120,7 @@ PendingAsyncCall Proxy::callMethod(const MethodCall& message, async_reply_handle
     SDBUS_THROW_ERROR_IF(!message.isValid(), "Invalid async method call message provided", EINVAL);
 
     auto callback = (void*)&Proxy::sdbus_async_reply_handler;
-    auto callData = std::make_shared<AsyncCalls::CallData>(AsyncCalls::CallData{*this, std::move(asyncReplyCallback), {}});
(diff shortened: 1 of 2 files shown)


sangelovic:
@dreiss Thanks for the description of the bug and for the fix!

sangelovic:
I made a little change -- `enum class` instead of an `int` -- for higher type safety

PR #376, opened 2023-11-05
Minor UnixFd cleanups

dreiss:


include/sdbus-c++/Types.h
```diff
@@ -237,6 +239,9 @@ namespace sdbus {
 
         UnixFd& operator=(UnixFd&& other)
         {
+            if (this == &other) {
+                return *this;
+            }
             close();
             fd_ = other.fd_;
             other.fd_ = -1;
```
dleeds-cpi, line 247:
```suggestion
            fd_ = std::exchange(other.fd_, -1);
```
Can also use this in `release()`.

include/sdbus-c++/Types.h
```diff
@@ -225,8 +224,11 @@ namespace sdbus {
 
         UnixFd& operator=(const UnixFd& other)
         {
+            if (this == &other) {
```
dleeds-cpi, line 227: Style convention appears to be:
```suggestion
            if (this == &other)
            {
```

src/Types.cpp
```diff
@@ -64,4 +67,24 @@ bool Variant::isEmpty() const
...
 }
 
+void UnixFd::close()
+{
+    if (fd_ >= 0) {
+        ::close(fd_);
+    }
+}
+
+int UnixFd::checked_dup(int fd)
```
dleeds-cpi, line 77: Style convention appears to be:
```suggestion
int UnixFd::checkedDup(int fd)
```

dleeds-cpi (commented):
Always good to have better error checking.

PR #388, opened 2023-12-06
refactor: improve Object vtable registration

### LINKED ISSUE #149: Create signalregister wrapper
This "bug" costed me more than one day of hard thinking... :facepalm:

I have a sdbus client and I want to wait for two signals, so I used `uponSignal`. The convinient API really makes it hard to misuse, but not impossible, because I forgot `proxy->finishRegistration()`.

I would love to see something like this (just a draft):

```
proxy->register_signals(
    sdbus::createSignal("com.example.iface", "Completed", [](int result) { fmt::print("I got {}\n", result); ),
    sdbus::createSignal("org.freedesktop.DBus.Properties", "PropertiesChanged", propchange_callback),
);
```

where in the end the `register_signals` function will call `finishRegistration` on its own.

sangelovic:
Fixes #149


hellow554:
Hey sangelovic,

amazing work! I really like that new API.
A few nitpicks:
Add a description in this PR message, at least say something like `closes #149` or similar. 
Also it's one big commit again (+486,-806) which is huge and not very reviewable. But I believe in the tests, so that is okay I think.

Amazing work, can't wait to use the API!

sangelovic:
@hellow554 Thanks, I will do that.

PR #393, opened 2023-12-30
refactor: use async tag to denote async calls or operations

sangelovic:


dleeds-cpi (commented):
@sangelovic are you sure this is a helpful change? I have a few concerns:
* Overloading synchronous and asynchronous variants makes two quite different operations look similar, which could result in misuse. For example, I want to use a try/catch for the sync variant and an error handler in the callback for the async variant.
* Function calls get longer and more cumbersome when switching from an `Async` suffix to a `sdbus::async` tag.
* Tab completion may become less useful for IDE users.
* This is a massive API change and will make it challenging to switch existing codebases to v2.0.

sangelovic:
@dleeds-cpi Thanks for your comment. The idea was long in my head to use implement async variants with tag dispatching to

* get more consistent with other tag-based solutions in the library API,
* stop mixing abstractions -- i.e. both *what* the function does and *how* it does it in a method name, and keep only the *what* part. Tag types are [ideal for that](https://www.fluentcpp.com/2018/04/27/tag-dispatching/).

E.g. `proxy->GetAsync(...)` on `org.freedesktop.DBus.Properties` interface may sound confusing at the first look (I would ask myself: what an "async" object are we getting here?), where `proxy->Get(..., async);` is clearer to my right at the first sight (where I see we get something in an async way). But perhaps this is a little contrived example given the not very well chosen method name (`GetPropertyAsync()` would probably cause less or no confusion), but I chose this because I wanted to adhere to the D-Bus specification here.

This solution is also a little more generic programming-friendly. But I'm not sure if sdbus-c++ API would ever be used in that context by anyone :-)

I admit the downsides. The function call will get a little longer. However, with imported namespace/name, it would be similar (`GetAsync()` vs `Get(async)`.

I also see that in some cases, where functions have more parameters, the `async` tag is "buried" far at the end of the argument list, so it may not be so obvious at the first sight.

Regarding your first bullet point, we already have the problem with `IProxy::callMethod()` overloads taking the method call message, where the sync and async variants have the same `callMethod()` name.

And yes, the change is massive for those using async variants...

You are probably using sdbus-c++ API quite a lot in an async setup, and have experience with async API's in general. If you think my two bullet points above are not worth it, I'll reconsider and potentially drop this PR.

But then, for consistency, I would suggest to introduce breaking `Async` suffix to the above-mentioned async message-based `callMethod()` overloads, for consistency.

What are your thoughts?

dleeds-cpi:
@sangelovic it is very much subjective. One the one hand, it could result in a cleaner looking API, although I have seen more people become confused by tag parameters than by variants being represented by a function name pattern. That said, I think people can get used to any pattern, as long as it is applied consistently.

Examples:
```c++
// Sync
try {
    proxy->callMethod("SetTimezone")
            .onInterface("org.freedesktop.timedate1")
            .withArguments("America/Los_Angeles", false);
} catch (const sdbus::Error &e) {
    // Error
}

// Async old
proxy->callMethodAsync("SetTimezone")
        .onInterface("org.freedesktop.timedate1")
        .withArguments("America/Los_Angeles", false)
        .uponReplyInvoke([value](const sdbus::Error *e) {
            if (e) {
                // Error
            } else {
                // Yay
            }
        });

// Async new
proxy->callMethod("SetTimezone", sdbus::async)
        .onInterface("org.freedesktop.timedate1")
        .withArguments("America/Los_Angeles", false)
        .uponReplyInvoke([value](const sdbus::Error *e) {
            if (e) {
                // Error
            } else {
                // Yay
            }
        });
```
The above examples seem clear enough on intent, so as long as users are aware of the async tagged overload, there should not be undue confusion.

PR #396, opened 2024-01-09
refactor: use optional for passing potential call errors

sangelovic:
This switches from a raw pointer to `std::optional` type to pass prospective call errors to the client (using `std::optional` was not possible years back when sdbus-c++ was based on C++14). This makes the API a little clearer, safer, idiomatically more expressive, and removes potential confusion associated with raw pointers (like ownership, lifetime questions, etc.). However, this is a *breaking* API change (for async users). The upcoming v2 release is a nice candidate to accommodate such a change.

@dleeds-cpi What are your thoughts about it?

### DIFF
--- a/docs/using-sdbus-c++.md
+++ b/docs/using-sdbus-c++.md
@@ -167,10 +167,10 @@ All public types and functions of sdbus-c++ reside in the `sdbus` namespace.
 Error signalling and propagation
 --------------------------------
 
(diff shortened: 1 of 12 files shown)


dleeds-cpi:
@sangelovic I like it! It definitely feels more modern, but does not present a significant change to usage patterns.

PR #398, opened 2024-01-17
Add hasProxy to ProxyObjectHolder

CMurtagh-LGTM:
The held proxy object can be moved out of the unique_ptr so it is possible to point to null but there is no way to inspect if this is the case.

### DIFF
--- a/include/sdbus-c++/ProxyInterfaces.h
+++ b/include/sdbus-c++/ProxyInterfaces.h
@@ -69,6 +69,11 @@ namespace sdbus {
             return *proxy_;
         }
 
+        bool hasProxy()
+        {
+            return proxy_ == nullptr;
+        }
+
     private:
         std::unique_ptr<IProxy> proxy_;
     };


sangelovic:
A moved-from proxy or moved-from object (i.e., an instance of `ProxyObjectHolder`, `ObjectHolder`, and all derived classes down the inheritance hierarchy) in sdbus-c++ is an object in an **invalid** state. Further accessing the object or invoking operations (methods) on the object is undefined behavior. Hence, no querying is needed to see if an object is valid or invalid. A moved-from one is simply always invalid.

Since you explicitly call `std::move` to move an object, you know that object and you should not be surprised by it (and the compilers do implicit moves only when it's not possible to do anything to the moved-from object).

Btw. upcoming sdbus-c++ v2 API will provide [methods](https://github.com/Kistler-Group/sdbus-cpp/blob/release/v2.0/include/sdbus-c%2B%2B/AdaptorInterfaces.h#L138) to access the underlying `Proxy`/`Object` ptrs.

PR #409, opened 2024-02-19
chore(cmake): implement more flexible searching for sd-bus libs

### LINKED ISSUE #399: Manually configuring choice of sd-bus implementation library
While I appreciate that this library can work without systemd or elogind, there is still a problem. For example, when you have elogind and basu installed on your system simultaneously, there is no way for user to specify using basu - cmake will always choose elogind for him.

If I were an expert in cmake, I would add an option:
```sh
cmake -DIMPL=basu     # builds sdbus-c++ with basu, ignoring presence of systemd & elogind
cmake -DIMPL=systemd  # builds sdbus-c++ with systemd, ignoring presence of elogind & basu
cmake -DIMPL=elogind  # builds sdbus-c++ with elogind, ignoring presence of systemd & basu
cmake -DIMPL=something_else  # configuration error
cmake  # builds with what is present on a system
```
(or maybe another name instead of `IMPL`)

sangelovic:
A new CMake configuration variable `SDBUSCPP_SDBUS_LIB` has been introduced that enables clients to specify which sd-bus implementation library shall be used by sdbus-c++.

Fixes #399

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,19 +12,22 @@ include(GNUInstallDirs) # Installation directories for `install` command and pkg
 # CONFIGURATION OPTIONS
 # -------------------------------
 
-option(SDBUSCPP_BUILD_LIBSYSTEMD "Fetch & build libsystemd static library and make it part of libsdbus-c++, instead of searching for libsystemd in the system" OFF)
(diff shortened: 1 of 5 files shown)


lch361:
Nice!

PR #415, opened 2024-02-29
Support serialization and deserialization of std::variant

anthonybrandon:
It can be useful to directly serialize from and deserialize to `std::variant<>` instead of `sdbus::Variant`.

### DIFF
--- a/docs/using-sdbus-c++.md
+++ b/docs/using-sdbus-c++.md
@@ -1516,7 +1516,7 @@ sdbus-c++ provides many default, pre-defined C++ type representations for D-Bus 
 | string-like, basic  | 103         | g          | SIGNATURE          | `sdbus::Signature`                |
 | container           | 97          | a          | ARRAY              | `std::vector<T>`, `std::array<T>`, `std::span<T>` - if used as an array followed by a single complete type `T` <br /> `std::map<T1, T2>`, `std::unordered_map<T1, T2>` - if used as an array of dict entries |
 | container           | 114,40,41   | r()        | STRUCT             | `sdbus::Struct<T1, T2, ...>` variadic class template                               |
-| container           | 118         | v          | VARIANT            | `sdbus::Variant`                  |
+| container           | 118         | v          | VARIANT            | `sdbus::Variant`, `std::variant<T1, ...>` |
 | container           | 101,123,125 | e{}        | DICT_ENTRY         | -                               |
 | fixed, basic        | 104         | h          | UNIX_FD            | `sdbus::UnixFd`                   |
 | reserved            | 109         | m          | (reserved)         | -                               |
--- a/include/sdbus-c++/Message.h
+++ b/include/sdbus-c++/Message.h
@@ -96,6 +96,9 @@ namespace sdbus {
         Message& operator<<(const std::vector<_Element, _Allocator>& items);
         template <typename _Element, std::size_t _Size>
         Message& operator<<(const std::array<_Element, _Size>& items);
+
+        template <typename ...Elements>
+        Message& operator<<(const std::variant<Elements...>& value);
 #if __cplusplus >= 202002L
         template <typename _Element, std::size_t _Extent>
         Message& operator<<(const std::span<_Element, _Extent>& items);
@@ -128,6 +131,8 @@ namespace sdbus {
         Message& operator>>(std::vector<_Element, _Allocator>& items);
         template <typename _Element, std::size_t _Size>
         Message& operator>>(std::array<_Element, _Size>& items);
+        template <typename ...Elements>
+        Message& operator>>(std::variant<Elements...>& value);
 #if __cplusplus >= 202002L
         template <typename _Element, std::size_t _Extent>
         Message& operator>>(std::span<_Element, _Extent>& items);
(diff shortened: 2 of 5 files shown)


sangelovic:
Thanks for the PR, it's interesting and useful. I edited your code directly -- I've done some minor adjustments (formatting mostly, for consistency with existing code), and I've also simplified slightly the deserialization template so its invocation is simpler.

Out of curiosity, I also played with some implementation variations of that deserialization template. One alternative would be:

```c++
namespace detail
{
	template <typename _Element, typename _Variant>
	bool deserialize_variant(Message& msg, _Variant& value, const std::string& signature)
	{
		if (signature != signature_of<_Element>::str())
			return false;

		_Element temp;
		msg.enterVariant(signature);
		msg >> temp;
		msg.exitVariant();
		value = std::move(temp);
		return true;
	}

	template <typename _Variant, std::size_t... _Is>
	void deserialize_variant(Message& msg, _Variant& value, const std::string& signature, std::index_sequence<_Is...>)
	{
		bool result = (!deserialize_variant<std::variant_alternative_t<_Is, _Variant>>(msg, value, signature) || ...);
		SDBUS_THROW_ERROR_IF(!result, "Failed to deserialize variant: signature did not match any of the variant types", EINVAL);
	}
}

template <typename... _Elements>
inline Message& Message::operator>>(std::variant<_Elements...>& value)
{
	std::string type;
	std::string contentType;
	peekType(type, contentType);
	detail::deserialize_variant(*this, value, contentType, std::index_sequence_for<_Elements...>{});
	return *this;
}
```

I like that this one is neat and compact. With C++20 we could use template lambda instead of `bool deserialize_variant(...)`. But perhaps it's more cryptic to an average developer...

The other would be:

```c++
namespace detail
{
	template <std::size_t _I = 0, typename... _Elements>
	void deserialize_variant(Message& msg, std::variant<_Elements...>& value, const std::string& signature)
	{
		if constexpr (_I < std::variant_size_v<std::variant<_Elements...>>)
		{
			using Element = std::variant_alternative_t<_I, std::variant<_Elements...>>;
			if (signature == signature_of<Element>::str())
			{
				Element temp;
				msg.enterVariant(signature);
				msg >> temp;
				msg.exitVariant();
				value = std::move(temp);
			}
			else
				deserialize_variant2<_I + 1>(msg, value, signature);
		}
		else
			SDBUS_THROW_ERROR("Failed to deserialize variant: signature did not match any of the variant types", EINVAL);
	}
}

template <typename... _Elements>
inline Message& Message::operator>>(std::variant<_Elements...>& value)
{
	std::string type;
	std::string contentType;
	peekType(type, contentType);
	detail::deserialize_variant(*this, value, contentType);
	return *this;
}
```

Here I like that we have one function only, and the invocation is also very simple, and it's perhaps more readable and expressive what the code does (recursive "iteration" with incrementing index...)

Which one of the three do you prefer most? I've kept the original solution of yours for now, just simplified a bit as mentioned above.

anthonybrandon:
Sorry about that, the coding style for my current project made its way into this PR.
I like the first option you gave, but I wonder if there is a reason to use the index_sequence?
We could do something along these lines:

```cpp
    namespace detail
    {
        template <typename _Element, typename ..._Elements>
        bool deserialize_variant(Message& msg, std::variant<_Elements...>& value, const std::string& signature)
        {
            if (signature != signature_of<_Element>::str())
                return false;

            _Element temp;
            msg.enterVariant(signature);
            msg >> temp;
            msg.exitVariant();
            value = std::move(temp);
            return true;
        }
    }

    template <typename ..._Element>
    inline Message& Message::operator>>(std::variant<_Element...>& value)
    {
        std::string type;
        std::string contentType;
        peekType(type, contentType);
        bool result = (detail::deserialize_variant<_Element, _Element...>(*this, value, contentType) || ...);
        SDBUS_THROW_ERROR_IF(!result, "Failed to deserialize variant: signature did not match any of the variant types", EINVAL);
        return *this;
    }
    ```

What do you think? Now there is also only one function and both seem pretty simple to me.
The only part I don't like is having to repeat the `_Elemen, Element...` at the call site.

I also have a commit adding `std::variant` to the table of supported types. Should I add it to this PR?

**Comment by sangelovic** [comment]
@anthonybrandon Oh right, of course, we can do the unfolding directly on `_Elements...`, no need for the `index_sequence` here, good point! In addition, there is no need to explicitly mention `_Elements...` template parameter at the call site, because the `_Elements...` template argument will be deduced automatically from the variant function parameter:

```c++
namespace detail
{
    template <typename _Element, typename... _Elements>
    bool deserialize_variant(Message& msg, std::variant<_Elements...>& value, const std::string& signature)
    {
        if (signature != signature_of<_Element>::str())
            return false;

        _Element temp;
        msg.enterVariant(signature);
        msg >> temp;
        msg.exitVariant();
        value = std::move(temp);
        return true;
    }
}

template <typename... Elements>
inline Message& Message::operator>>(std::variant<Elements...>& value)
{
    std::string type;
    std::string contentType;
    peekType(type, contentType);
    bool result = (detail::deserialize_variant<Elements>(*this, value, contentType) || ...);
    SDBUS_THROW_ERROR_IF(!result, "Failed to deserialize variant: signature did not match any of the variant types", EINVAL);
    return *this;
}
```

With this I think we have a nice, acceptable solution, don't we? Feel free to update the MR, and please can you also add `std::variant` to the table of supported types... Thanks.

anthonybrandon:
Okay, right this looks good. I'll update the MR. For some reason I thought you had to specify all template parameters if you specified any.

sangelovic:
Looks good, thank you, will merge it.

One additional thought... The generated bindings use `sdbus::Variant` because it's a type erased class, whereas `std::variant`'s internal types form its class type and the binding generator has no idea what these types might be. Would it make sense if we add a converting constructor of `sdbus::Variant` from `std::variant` (and, conversely, a conversion operator from `sdbus::Variant` to `std::variant`)? It could simplify the interoperability between these two D-Bus variant representations. What do you think?

anthonybrandon:
I think it would make sense. I can implement it if you like.

sangelovic:
Of course, looking forward to your PR ;o)

-------- Pôvodná správa --------
1. 4. 2024 13:29, Anthony Brandon  napísal/a:

PR #426, opened 2024-04-01
Fix signature_of for std::variant

anthonybrandon:
Currently all std::variants are treated as if they have a valid signature but only variants whose elements have a valid signature should be treated as having a valid signature.

### DIFF
--- a/include/sdbus-c++/TypeTraits.h
+++ b/include/sdbus-c++/TypeTraits.h
@@ -332,8 +332,14 @@ namespace sdbus {
         }
     };
 
+    template <typename T>
+    using signature_valid = std::bool_constant<signature_of<T>::is_valid>;
+
+    template <typename... T>
+    using signatures_valid = std::conjunction<signature_valid<T>...>;
+
     template <typename... Elements>
-    struct signature_of<std::variant<Elements...>> : signature_of<Variant>
+    struct signature_of<std::variant<Elements...>, std::enable_if_t<signatures_valid<Elements...>::value>> : signature_of<Variant>
     {};
 
     template <>


anthonybrandon:
@sangelovic While trying to implement the std::variant conversion operator I realized it doesn't make sense for std::variant to have a valid signature if not all its elements do.
Please check if the resulting compiler errors are okay, or if you prefer to have the std::enable_if somewhere else in order
to get different compiler errors. I guess it's not possible to have tests for non-compiling code.

On a somewhat related note I found that `sdbus::Variant` has the following operator:

```
        // Only allow conversion operator for true D-Bus type representations in C++
        template <typename _ValueType, typename = std::enable_if_t<signature_of<_ValueType>::is_valid>>
        operator _ValueType() const
        {
            return get<_ValueType>();
        }
```

This somewhat conflicts with adding an operator for std::variant<>, since then it's not clear which behavior is expected.
Do we expect to extract a variant from within the sdbus::Variant (i.e. `std::variant<std::variant<>>`, or do we expect to extract a non variant value from the variant. Currently `operator _ValueType` will attempt to extract a variant from within the variant, which is consistent with the behavior for other types, though possibly not expected.
So it will call `msg_.enterVariant("v");` Should I still add `operator std::variant`?

sangelovic:
As for the first point, compiler error messages are more targeted and point specifically at the unsupported type within variant. While error messages coming from excluding the variant as a whole when at least one if its internal types is unsupported would be confusing to me as a client. So I would rather not introduce the change in this PR in this respect.

anthonybrandon:
okay, then I will close this.

sangelovic:
I'm looking into the second point you brought up... will come back soon

sangelovic:
> So it will call msg_.enterVariant("v"); Should I still add operator std::variant?

Good point. I got unsure if underlying sd-bus library in its serialization implementation allows variants of variants. I checked -- and it allows them (variants are simply containers which can host elements which themselves are containers, thus variants, too...). They are also possible with low-level sdbus-c++ API (`Message`-based API). However, high-level sdbus-c++ API does not allow that. One cannot create a `sdbus::Variant{sdbus::Variant{aString}}`, because this calls the copy constructor.

Frankly, I have never seen a use case of variants of variants, and I have difficulties imagining its usefulness. Looks like sdbus-c++ clients never needed that either, for no one came with an issue or PR for that.

So I would tend to be conservative here and keep that approach. Now that we've added `std::variant`, we actually broke this concept, because now one can create a variant of variant through `sdbus::Variant{std::variant<int, std::string>{aString}}`. Now, we in fact need to explicitly define the converting constructor `sdbus::Variant(const std::variant<...>& v);` in order to fix that and to be consistent.

This, symmetrically, also answers the question about the conversion operator. Yes, we need to explicitly define the conversion operator from `sdbus::Variant` to `std::variant`.

What I would imagine is that in the end `sdbus::Variant` and `std::variant` are inter-changeable, like two containers. They represent the same thing and if we convert one to the other, the content of the former shall exactly be the content of the latter.

What do you think?

sangelovic:
> Because sdbus::Variant has operator bool. I'm not sure how to fix that.

I guess this happens as a side effect of SFINAE, where the `std::variant`-based operator template instantiation leads to an error (because `FooBar` is not a supported D-Bus type representation and hence either `signature_of` or `Message::operator<<` fails), so the compiler picks the other operator template and tries to construct the variant element-wise. It only happens when one of the variant alternatives is unsupported...

In my view, nothing helps here except of marking the `operator _ValueType()` as explicit (whereas the `std::variant` one can remain implicit). This actually:

*  is a good practice according to the C++ core guidelines, as it makes the code safer and better follows the principle of least astonishment
*  is consistent with the standard library style, where wrappers like `std::variant`, `std::any`, `std::optional`... also do not provide an implicit conversion to the underlying type.

**But...** such a change would be a breaking change. Hence I (already) prepared it for the next major release: https://github.com/Kistler-Group/sdbus-cpp/pull/428, and I suggest we keep it as is in current master (`v1.x`) line.

anthonybrandon:
```
What I would imagine is that in the end sdbus::Variant and std::variant are inter-changeable, like two containers. They represent the same thing and if we convert one to the other, the content of the former shall exactly be the content of the latter.

What do you think?
```

Yes, this makes the most sense to me.

```
But... such a change would be a breaking change. Hence I (already) prepared it for the next major release: https://github.com/Kistler-Group/sdbus-cpp/pull/428, and I suggest we keep it as is in current master (v1.x) line.
```

Yes, this was the first thing I tried, but it breaks a lot of existing code so figured it probably wasn't desirable.

sangelovic:
OK. So you will prepare a PR according to our conclusions, right?

anthonybrandon:
Yes I will prepare it.

PR #429, opened 2024-04-02
Add std::variant constructor and conversion operator to sdbus::Variant

anthonybrandon:


tests/unittests/Types_test.cpp
```diff
@@ -130,6 +138,16 @@ TEST(AVariant, ReturnsTrueWhenAskedIfItContainsTheTypeItReallyContains)
...
 }
 
+TEST(AVariant, CanBeConvertedIntoAnStdVariant)
+{
+    using VariantType = std::variant<std::string, uint64_t, std::vector<sdbus::Struct<std::string, double>>>;
+    VariantType value{ std::vector<sdbus::Struct<std::string, double>>{{"hello"s, ANY_DOUBLE}, {"world"s, ANY_DOUBLE}} };
+
+    sdbus::Variant variant(value);
+
+    ASSERT_EQ(value, VariantType { variant });
```
sangelovic, line 148: Would you please change to:

```suggestion
    ASSERT_THAT(value, Eq(VariantType{variant}));
```

We use this Hamcrest notation in sdbus-c++ tests as we find it more readable and clear what is being tested against what (the only exception is testing simply boolean expressions, using `ASSERT_TRUE()`/`ASSERT_FALSE()`)

Also minor, but I would rename `VariantType` to `StdVariantType` for better clarity...

tests/unittests/Types_test.cpp
```diff
@@ -130,6 +138,16 @@ TEST(AVariant, ReturnsTrueWhenAskedIfItContainsTheTypeItReallyContains)
...
+}
+
```
sangelovic, line 150: Thinking more about it, I would suggest we split this test in two. One would test and demonstrate that we can do the conversion, the other would test and demonstrate that the two variant types are implicitly interchangeable (given my first comment above):

```suggestion
TEST(AVariant, CanBeConvertedIntoAnStdVariant)
{
    std::vector<sdbus::Struct<std::string, double>> value{{"hello"s, ANY_DOUBLE}, {"world"s, ANY_DOUBLE}};
    using StdVariantType = std::variant<std::string, uint64_t, std::vector<sdbus::Struct<std::string, double>>>;
    sdbus::Variant sdbusVariant(value);

    StdVariantType stdVariant{sdbusVariant};

    ASSERT_TRUE(std::holds_alternative<std::vector<sdbus::Struct<std::string, double>>>(stdVariant));
    ASSERT_THAT(std::get<std::vector<sdbus::Struct<std::string, double>>>(stdVariant), Eq(value));
}

TEST(AVariant, IsImplicitlyInterchangeableWithStdVariant)
{
    std::vector<sdbus::Struct<std::string, double>> value{{"hello"s, ANY_DOUBLE}, {"world"s, ANY_DOUBLE}};
    using StdVariantType = std::variant<std::string, uint64_t, std::vector<sdbus::Struct<std::string, double>>>;
    StdVariantType stdVariant{value};

    auto stdVariantCopy = [](const sdbus::Variant& v) -> StdVariantType{ return v; }(stdVariant);

    ASSERT_THAT(stdVariantCopy, Eq(stdVariant));
}

```

tests/unittests/Types_test.cpp
```diff
@@ -72,6 +72,14 @@ TEST(AVariant, CanBeConstructedFromAComplexValue)
     ASSERT_NO_THROW(sdbus::Variant{value});
 }
 
+TEST(AVariant, CanBeConstructedFromAnStdVariant)
+{
+    using VariantType = std::variant<std::string, uint64_t, std::vector<sdbus::Struct<std::string, double>>>;
+    VariantType value{ std::vector<sdbus::Struct<std::string, double>>{{"hello"s, ANY_DOUBLE}, {"world"s, ANY_DOUBLE}} };
+
+    ASSERT_NO_THROW(sdbus::Variant{value});
+}
```
sangelovic, line 81: I would make it stricter and express explicitly that the resulting `sdbus::Variant` does not contain the `value`, but the underlying value of `value`:

```suggestion
TEST(AVariant, CanBeConstructedFromAnStdVariant)
{
    std::vector<sdbus::Struct<std::string, double>> value{{"hello"s, ANY_DOUBLE}, {"world"s, ANY_DOUBLE}};
    using StdVariantType = std::variant<std::string, uint64_t, std::vector<sdbus::Struct<std::string, double>>>;
    StdVariantType stdVariant{value};

    sdbus::Variant sdbusVariant(value);

    ASSERT_TRUE(variant.containsValueOfType<std::vector<sdbus::Struct<std::string, double>>>());
    ASSERT_THAT(variant.get<std::vector<sdbus::Struct<std::string, double>>>, Eq(value));
}
```

What do you think?

sangelovic (commented):
Thanks for the PR, well done... I added a couple of suggestions for discussion.

include/sdbus-c++/Types.h
```diff
@@ -65,6 +65,14 @@ namespace sdbus {
             msg_.seal();
         }
 
+        template <typename... _Elements>
+        explicit Variant(const std::variant<_Elements...>& value)
```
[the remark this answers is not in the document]
anthonybrandon: Okay, for me it doesn't matter. I will change it.

tests/unittests/Types_test.cpp
anthonybrandon: I assume here you mean `sdbus::Variant sdbusVariant{stdVariant}`?
Otherwise we aren't testing the new constructor.

anthonybrandon: I think the tests are better this way compared to what I had before.

sangelovic: > I assume here you mean `sdbus::Variant sdbusVariant{stdVariant}`? Otherwise we aren't testing the new constructor.

Yes, that's what I meant actually.

> I wonder if we need the extra ASSERT_TRUE.

Not from the verification perspective. I kept it as an explicit statement for future readers (since tests are an up-to-date documentation of behaviors) that the type underneath the variant is vector, not variant. But feel free to remove it if you consider that to be clear from the second assert.

anthonybrandon: No if it's just to be more explicit about the intention I think it's fine to keep it.

PR #440, opened 2024-04-30
feat: add macro to teach sdbus-c++ about user-defined structs

sangelovic:
This introduces `SDBUSCPP_REGISTER_STRUCT()` macro that helps clients conveniently, in one line, teach sdbus-c++ to recognize and accept their custom C++ struct types wherever D-Bus structs are expected in a D-Bus API.

The macro saves some boilerplate that would otherwise be needed on client side for every registered struct.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,19 +14,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04]
-        compiler: [g++]
-        build: [shared-libsystemd]
+        os: [ubuntu-22.04]
+        compiler: [g++, clang]
+        build: [shared-libsystemd, embedded-static-libsystemd]
         include:
-          - os: ubuntu-22.04
-            compiler: clang
+          - os: ubuntu-20.04
+            compiler: gcc
             build: shared-libsystemd
-          - os: ubuntu-22.04
-            compiler: g++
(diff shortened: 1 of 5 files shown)


sangelovic:
@hellow554 I prepared a macro that avoids the boilerplate and registers a struct in sdbus-c++ as a one-liner. E.g. `SDBUSCPP_REGISTER_STRUCT(PersonData, name, surname, age, address);`. Have a look into this MR. Does it look attractive to you? :-)

It supports up to 16 struct members currently. Is that enough for you?

hellow554:
That really looks good. Thanks a lot for that :)

> It supports up to 16 struct members currently. Is that enough for you?

I guess?
You could add something like this maybe?

```
#define SDBUSCPP_REGISTER_STRUCT(struct, ...)  \
  #error "Not more than 16 Parameters are supported, please open an issue if you would like to see this feature"
```

Not sure if that's possible thought...

sangelovic:
@hellow554 It's not possible AFAIK. I've added a static assert that checks that the count of arguments if less or equal to 16, but it only works nicely (i.e. the compilation stops there showing directly the accompanying message) if there is not more than 32 arguments. Otherwise the number expands to one of the struct member names, leading to wrong syntax, which stops the compilation with a slightly different message, but at least in my compiler the 3rd link in the error message points to the static assert code (because that's where there is an undefined symbol), so after some digging clients can figure out the cause of the compilation failure.

PR #441, opened 2024-05-10
Update using-sdbus-c++.md

lctasca:
Update in using documentation to version 2 syntax making examples work out of the box


sangelovic:
Thanks for the fixes!

Issue #460, opened 2024-10-24
Add --version flag to xml2cpp to print out sdbus version


br337:
**Is your feature request related to a problem? Please describe.**
Since the output generated by the xml2cpp tool is dependent on the version of sdbus the tool has been built with, it would be nice to provide a way to print out the sdbus version for which the xml2cpp tool is working. 

**Describe the solution you'd like**
A --version flag to display the sdbus version for which the xml2cpp tool is generating code.

**Describe alternatives you've considered**
Not many alternatives other than hoping cmake finds the correct xml2cpp binary, while also making sure that multiple versions aren't installed on the host.

**Additional context**
Due to some limitations (cross compiling code and using the natively built xml2cpp tool as a build time dependency to generate the necessary adapter code) the xml2cpp can't easily be included as a cmake project using the find_package directive without building it from source and thereby having to build libexpat from source as well. This can be quite a hassle to do in a cross compilation environment and not strictly necessary. Thereby, the find_program cmake directive is used to lookup the xml2cpp tool. At the same time, one can't always be sure which exact xml2cpp binary version cmake has found. It would be helpful to have a --version flag that outputs the currently running version of the xml2cpp tool so that one can verify that the correct xml2cpp tool has been found by cmake during the project configuration.

br337:
I could implement this myself and create a PR if the maintainers are okay with this. My solution would be a cmake generated header file that sets a define with the current version taken from the project version (by looking at the xml2cpp tool it seems that the tool's cmake project version follows the sdbus' library version). The xml2cpp tool would then in turn print out the defined value when passed the --version flag.

sangelovic:
Hey @br337, yes, please open a PR with the solution you proposed (an alternative that comes to my mind would be to pass the version as a preprocessor definition via `target_compile_definitions()`, but I'll leave the final choice up to you). Thank you.

PR #463, opened 2024-11-03
Added -v, --version command line tool options to xml2cpp 

br337:
Adds the `-v, --version` command line tool options to the `xml2cpp` tool.

Closes https://github.com/Kistler-Group/sdbus-cpp/issues/460 

Usage:
```bash
$ ./sdbus-c++-xml2cpp --version
Version: 2.0.0
```

```bash
$ ./sdbus-c++-xml2cpp -v
Version: 2.0.0
```

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -6,8 +6,6 @@ cmake_minimum_required(VERSION 3.5)
 
 project(sdbus-c++-tools VERSION 2.0.0)
 
-set(sdbus-c++-tools_VERSION ${CMAKE_PROJECT_VERSION})
-
 include(GNUInstallDirs)
 
 #-------------------------------
@@ -46,7 +44,7 @@ set(CMAKE_CXX_STANDARD 14)
 add_executable(sdbus-c++-xml2cpp ${SDBUSCPP_XML2CPP_SRCS})
 target_link_libraries (sdbus-c++-xml2cpp ${EXPAT_LIBRARIES})
 target_include_directories(sdbus-c++-xml2cpp PRIVATE ${EXPAT_INCLUDE_DIRS})
-target_compile_definitions(sdbus-c++-xml2cpp PRIVATE SDBUS_XML2CPP_VERSION="${sdbus-c++-tools_VERSION}")
+target_compile_definitions(sdbus-c++-xml2cpp PRIVATE SDBUS_XML2CPP_VERSION="${CMAKE_PROJECT_VERSION}")
 
 #----------------------------------
 # CMAKE CONFIG & PACKAGE CONFIG
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -6,6 +6,8 @@ cmake_minimum_required(VERSION 3.5)
 
 project(sdbus-c++-tools VERSION 2.0.0)
 
+set(sdbus-c++-tools_VERSION ${CMAKE_PROJECT_VERSION})
+
 include(GNUInstallDirs)
 
 #-------------------------------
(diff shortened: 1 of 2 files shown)


tools/xml2cpp-codegen/xml2cpp.cpp
```diff
@@ -51,6 +51,7 @@ void usage(std::ostream& output, const char* programName)
             "      --adaptor=FILE   Generate header file FILE with stub class (server)" << endl <<
             "  -h, --help           " << endl <<
             "      --verbose        Explain what is being done" << endl <<
+            "  -v, --version        Prints out the current SDBus version used by the tool" << endl <<
```
sangelovic, line 54:
```suggestion
            "  -v, --version        Prints out sdbus-c++ version used by the tool" << endl <<
```

tools/CMakeLists.txt
```diff
@@ -44,6 +46,7 @@ set(CMAKE_CXX_STANDARD 14)
 add_executable(sdbus-c++-xml2cpp ${SDBUSCPP_XML2CPP_SRCS})
 target_link_libraries (sdbus-c++-xml2cpp ${EXPAT_LIBRARIES})
 target_include_directories(sdbus-c++-xml2cpp PRIVATE ${EXPAT_INCLUDE_DIRS})
+target_compile_definitions(sdbus-c++-xml2cpp PRIVATE SDBUS_XML2CPP_VERSION="${sdbus-c++-tools_VERSION}")
```
sangelovic, line 49: Can't we use ${CMAKE_PROJECT_VERSION} here directly?

```suggestion
target_compile_definitions(sdbus-c++-xml2cpp PRIVATE SDBUS_XML2CPP_VERSION="${CMAKE_PROJECT_VERSION}")
```

Then we can drop unnecessary `sdbus-c++-tools_VERSION` variable.

br337: Reading on `CMAKE_PROJECT_VERSION` in [the cmake documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_PROJECT_VERSION.html) I found the following line:
```
In the event that the top level CMakeLists.txt contains multiple [project()](https://cmake.org/cmake/help/latest/command/project.html#command:project) calls, the most recently called one from that top level CMakeLists.txt will determine the value that CMAKE_PROJECT_VERSION contains.
```

So I decided to first grab the variable right after the project definition. If you consider this unnecessary then I can update the code accordingly.

sangelovic: Yes, this is unnecessary, since there are no other `project()` calls in between the latest `project()` call and `target_compile_definitions()`. Please update as suggested...

br337: Done

sangelovic: The is twice the dollar sign -- this is not intended, right?

tools/CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ set(CMAKE_CXX_STANDARD 14)
 add_executable(sdbus-c++-xml2cpp ${SDBUSCPP_XML2CPP_SRCS})
 target_link_libraries (sdbus-c++-xml2cpp ${EXPAT_LIBRARIES})
 target_include_directories(sdbus-c++-xml2cpp PRIVATE ${EXPAT_INCLUDE_DIRS})
+target_compile_definitions(sdbus-c++-xml2cpp PRIVATE SDBUS_XML2CPP_VERSION="$${CMAKE_PROJECT_VERSION}")
```
sangelovic, line 47: The dollar sign is twice here

br337: Amended the commit.

sangelovic: No problem, thanks for fixing :-)

sangelovic (approved):

PR #475, opened 2025-01-07
Using std::string::string

georgthegreat:
The following code utilizes class ctor, not `operator=` and was broken by sdbus-cpp 2.0.0:

```
const sdbus::ObjectPath SIGNALQ_DBUS_NOTIFY_MGR_OBJ_PATH = "/signalq/notify_mgr";
```

This fixes the behavior.

### DIFF
--- a/include/sdbus-c++/Types.h
+++ b/include/sdbus-c++/Types.h
@@ -194,14 +194,7 @@ namespace sdbus {
     class ObjectPath : public std::string
     {
     public:
-        ObjectPath() = default;
-        explicit ObjectPath(std::string value)
-            : std::string(std::move(value))
-        {}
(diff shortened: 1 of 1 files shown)


sangelovic:
This was intentional breaking change. `ObjectPath` is a strong type and can only be explicitly constructed from a `std::string` or C string, e.g. like this: `const sdbus::ObjectPath SIGNALQ_DBUS_NOTIFY_MGR_OBJ_PATH{"/signalq/notify_mgr"};`.

I guess you realized that yourself after opening the MR and closed the MR subsequently :-)

georgthegreat:
I realized that this PR breaked the build though I did not find the way to fix it.

Issue #490, opened 2025-04-23
Add `sd_bus_message_get_cookie()` and `sd_bus_message_get_reply_cookie()`


howey:
Use `sd_bus_message_get_cookie()` and `sd_bus_message_get_reply_cookie()` to get some unique identifier at the time an asynchronous method call is sent which can then be compared at the response to distinguish calls of the same method.

sangelovic:
Solved with #491

PR #491, opened 2025-04-23
feat: Add sd_bus_message_get_cookie (#490)

howey:
Add `sd_bus_message_get_cookie()` and `sd_bus_message_get_reply_cookie()` as `Message::getCookie()` and `Message::getReplyCookie()`.

### DIFF
--- a/include/sdbus-c++/Message.h
+++ b/include/sdbus-c++/Message.h
@@ -215,6 +215,7 @@ namespace sdbus {
         const char* getSender() const;
         const char* getPath() const;
         const char* getDestination() const;
+        uint64_t getCookie() const;
         // TODO: short docs in whole Message API
         std::pair<char, const char*> peekType() const;
         bool isValid() const;
@@ -296,6 +297,7 @@ namespace sdbus {
     public:
         MethodReply() = default;
         void send() const;
+        uint64_t getReplyCookie() const;
     };
 
     class Signal : public Message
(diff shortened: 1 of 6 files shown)


include/sdbus-c++/Message.h
```diff
@@ -215,6 +215,8 @@ namespace sdbus {
         const char* getSender() const;
         const char* getPath() const;
         const char* getDestination() const;
+        uint64_t getCookie() const;
+        uint64_t getReplyCookie() const;
```
sangelovic, line 219: `getReplyCookie()` only makes sense for method reply messages, right..? So then we should define it in `MethodReply` class, not in `Message`, what do you think?

howey: Sure

sangelovic:
I'm also thinking about a new (integration) test case verifying this feature and showing how the serial number in a method call message would match a serial number in the reply message. Not sure how familiar are you with the tests... I may help you here by writing it and pushing to your branch.

howey:
Sure, go ahead. This is roughly my code

```
  auto callback = [](sdbus::MethodReply reply, std::optional<sdbus::Error> error) {
    auto n = reply.getReplyCookie();
    std::cerr << "reply cookie " << n << std::endl;
  };

  for (int i = 0; i < 3; i++) {
    auto method = proxy->createMethodCall(interfaceName, methodName);
    proxy->callMethodAsync(method, callback);

    auto n = method.getCookie();
    std::cerr << "method cookie " << n << std::endl;
  }
```

sangelovic:
OK thanks, will do tomorrow.

sangelovic:
Sorry, took me longer than anticipated :-) But now the test is there, the CI is passing, so let's go ;-)

PR #501, opened 2025-07-08
Make it possible to use Result<void>

csk-ableton:
Makes it possible to use `sdbus::Result<void>` to bind async methods that don't return a value


csk-ableton:
> Why can't you simply use `sdbus::Result<>`? It's completely legal... and an empty type list idiomatically expresses that nothing is returned...

Oh it simply didn't occur to me, of course that is legal. Sorry for the noise.

PR #517, opened 2025-11-05
fix: create sdbus::Error correctly when errno is 0

### LINKED ISSUE #516: Defensive check for sd_bus_error_set_errno(…, 0) to avoid segfaults on empty error name
**Describe the bug**

Calling `sdbus::Error createError(int errNo, std::string customMsg)` from [`src/Error.cpp`](https://github.com/Kistler-Group/sdbus-cpp/blob/master/src/Error.cpp) with an errno value of 0 causes an empty error name to be propagated.

According to the [Debian sd_bus_error_set_errno manpage](https://manpages.debian.org/testing/libsystemd-dev/sd_bus_error_set_errno.3.en.html):
> sd_bus_error_set_errno() will immediately return 0 if the specified error parameter error is 0. Otherwise, it will set name from an errno-like value that is converted to a D-Bus error. strerror_r(3) will be used to set message. Well-known D-Bus error names will be used for name if applicable, otherwise a name in the "System.Error." namespace will be generated. The sign of the specified error number is ignored and the absolute value is used implicitly. If the specified error error is non-zero, the call always returns a negative value, for convenient usage in return statements. This call might fail due to lack of memory, in which case an SD_BUS_ERROR_NO_MEMORY error is set instead, and -ENOMEM is returned.

Specifically, if errno == 0, the error name is left as null, which means that the resulting `sdbus::Error` object has an empty name. If this flows through to error reply handlers (e.g. `Connection::createErrorReplyMessage`), it may result in a crash or assertion failure, because a non-empty error name is required by D-Bus and `sd_bus_error_set`.

**To Reproduce**
- Call or throw an sdbus-cpp error via:
  ```c++
  throw sdbus::createError(0, "Should not happen");
  ```
- The empty error name propagates through the library and can cause a segfault, as observed in my case.

**Expected behavior**
The library should guard against this misuse, for example:
- By immediately returning or doing nothing if errno is 0 (mimicking the C API),
- Or by throwing a developer-friendly exception/warning when an error with errno 0 is about to be emitted, instead of letting the crash propagate downstream.

**Actual behavior**
Currently, an empty D-Bus error name is emitted, leading to a segfault or assertion failure (undefined behavior).

**Additional context**
See: [`src/Error.cpp`](https://github.com/Kistler-Group/sdbus-cpp/blob/master/src/Error.cpp)
See: [Debian manpage - sd_bus_error_set_errno](https://manpages.debian.org/testing/libsystemd-dev/sd_bus_error_set_errno.3.en.html)

sangelovic:
Create an empty (invalid) `sdbus::Error` instance when errno is 0, instead of segfaulting or throwing an exception by `std::string`. Improve robustness of `sdbus::createError()` in general. Add more tests.

Fixes #516

### DIFF
--- a/src/Error.cpp
+++ b/src/Error.cpp
@@ -32,17 +32,24 @@
 
 namespace sdbus
 {
-    sdbus::Error createError(int errNo, std::string customMsg)
+    Error createError(int errNo, std::string customMsg)
     {
         sd_bus_error sdbusError = SD_BUS_ERROR_NULL;
         sd_bus_error_set_errno(&sdbusError, errNo);
         SCOPE_EXIT{ sd_bus_error_free(&sdbusError); };
 
-        Error::Name name(sdbusError.name);
+        Error::Name name(sd_bus_error_is_set(&sdbusError) ? sdbusError.name : "");
         std::string message(std::move(customMsg));
-        message.append(" (");
-        message.append(sdbusError.message);
-        message.append(")");
+        if (!message.empty() && sdbusError.message != nullptr)
+        {
+            message.append(" (");
+            message.append(sdbusError.message);
+            message.append(")");
+        }
+        else if (sdbusError.message != nullptr)
+        {
+            message = sdbusError.message;
+        }
 
         return Error(std::move(name), std::move(message));
     }
--- a/include/sdbus-c++/Error.h
+++ b/include/sdbus-c++/Error.h
@@ -64,7 +64,7 @@ namespace sdbus {
         }
 
         Error(Name name, std::string message)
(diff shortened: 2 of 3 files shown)


src/Error.cpp
```diff
@@ -29,6 +29,7 @@
 #include "ScopeGuard.h"
 
 #include SDBUS_HEADER
+#include <cassert>
```
JesseRiemens: I don't think this extra include is required

sangelovic:
@JesseRiemens I addressed all your comments. Please re-review...

PR #527, opened 2026-03-01
feat: add coroutine support for client-side asynchronous calls

alexcani:
C++ 20 introduced coroutine support to the language, enabling writing asynchronous code using the `co_await`, `co_return` and `co_yield` operators. Similar to other languages, coroutines allow functions to suspend execution at certain points and later resume from where they left off, without blocking the calling thread.

Coroutines are a natural fit for any sort of IO-bound asynchronous operations, such as D-Bus communication.

This PR introduces native coroutine support in sdbus-c++ by the means of the `Awaitable<T>` class, a type that implements the C++20 awaitable protocol, allowing D-Bus method calls to be awaited in coroutines.

#### Core library
- New `Awaitable<T>` type that can be `co_await`ed, which suspends a running coroutine. When the result or error of a method call arrives, the coroutine is resumed and the result/error is returned.
Implementation is done in a thread-safe way, meaning there are no race conditions between the awaitable object returned by asynchronous calls and the callback invoked by the event loop upon arrival of a message. Naturally, it works in the single-threaded scenario as well, where the connection event loop may be driven externally and/or integrated in a full fledged coroutine runtime.
- Low-level API: new `callMethodAsync()` overloads accepting `with_awaitable_t` tag and returning `Awaitable<MethodReply>`.
- High-level API: new `getResultAsAwaitable` methods in the relevant high-level API helpers, covering methods and properties. StandardInterfaces.h classes have also been updated to expose awaitable-based methods.
- Integration tests for both low-level and high-level APIs.

#### Codegen
- Added support for generating awaitable-based async methods in the xml2cpp tool through new values for the `org.freedesktop.DBus.Method.Async.ClientImpl` and `org.freedesktop.DBus.Property.[Get/Set].Async.ClientImpl` annotations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,9 @@ tests/run-test-on-device.sh
 .settings
 *.log
 
+# vscode
+.vscode/
+
 #autotools
 sdbus-cpp.pc
 *Makefile
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,6 +164,7 @@ set(SDBUSCPP_HDR_SRCS
     ${SDBUSCPP_SOURCE_DIR}/ISdBus.h)
 
 set(SDBUSCPP_PUBLIC_HDRS
+    ${SDBUSCPP_INCLUDE_DIR}/Awaitable.h
     ${SDBUSCPP_INCLUDE_DIR}/ConvenienceApiClasses.h
     ${SDBUSCPP_INCLUDE_DIR}/ConvenienceApiClasses.inl
     ${SDBUSCPP_INCLUDE_DIR}/VTableItems.h
--- a/docs/using-sdbus-c++.md
+++ b/docs/using-sdbus-c++.md
@@ -759,7 +759,7 @@ private:
 
 Analogously to the adaptor classes described above, there is one proxy class generated for one interface in the XML IDL file. The class is de facto a proxy to the concrete single interface of a remote object. For each D-Bus signal there is a pure virtual member function whose body must be provided in a child class. For each method, there is a public function member that calls the method remotely.
 
-Generated proxy classes are not copyable and not moveable by design. One can create them on the heap and manage them in e.g. a `std::unique_ptr` if move semantics is needed (for example, when they are stored in a container). 
+Generated proxy classes are not copyable and not moveable by design. One can create them on the heap and manage them in e.g. a `std::unique_ptr` if move semantics is needed (for example, when they are stored in a container).
 
 ```cpp
 /*
@@ -1130,11 +1130,11 @@ For a real example of a server-side asynchronous D-Bus method, please look at sd
 Asynchronous client-side methods
 --------------------------------
 
-sdbus-c++ also supports asynchronous approach at the client (the proxy) side. With this approach, we can issue a D-Bus method call without blocking current thread's execution while waiting for the reply. We go on doing other things, and when the reply comes, either a given callback handler will be invoked within the context of the event loop thread, or a future object returned by the async call will be set the returned value.6
+sdbus-c++ also supports asynchronous approach at the client (the proxy) side. With this approach, we can issue a D-Bus method call without blocking current thread's execution while waiting for the reply. We go on doing other things, and when the reply comes, either a given callback handler will be invoked within the context of the event loop thread, a future object returned by the async call will be set the returned value, or (with C++20) an awaitable can be `co_await`ed in a coroutine.
 
 ### Lower-level API
 
-Considering the Concatenator example based on lower-level API, if we wanted to call `concatenate` in an async way, we have two options: We either pass a callback to the proxy when issuing the call, and that callback gets invoked when the reply arrives:
+Considering the Concatenator example based on lower-level API, if we wanted to call `concatenate` in an async way, we have several options. We can pass a callback to the proxy when issuing the call, and that callback gets invoked when the reply arrives:
 
 ```c++
 int main(int argc, char *argv[])
@@ -1204,6 +1204,18 @@ Another option is to use `std::future`-based overload of the `IProxy::callMethod
     }
 ```
 
+A third option, available with C++20, is to use `sdbus::with_awaitable` to get an `Awaitable<MethodReply>` that can be `co_await`ed in a coroutine:
+
+```c++
+    // In a coroutine context:
+    auto method = concatenatorProxy->createMethodCall(interfaceName, concatenate);
+    method << numbers << separator;
+    auto reply = co_await concatenatorProxy->callMethod(method, sdbus::with_awaitable);
+    std::string result;
+    reply >> result;
+    // If an error occurs, sdbus::Error is thrown when co_await completes
+```
+
 ### Convenience API
 
 On the convenience API level, the call statement starts with `callMethodAsync()`, and one option is to finish the statement with `uponReplyInvoke()` that takes a callback handler. The callback is a void-returning function that takes at least one argument: `std::optional<sdbus::Error>`. All subsequent arguments shall exactly reflect the D-Bus method output arguments. A concatenator example:
@@ -1264,6 +1276,18 @@ The future object will contain void for a void-returning D-Bus method, a single 
         ...
 ```
 
+A third option, available with C++20, is to finish the async call statement with `getResultAsAwaitable<ReturnTypes...>()`, which returns an `Awaitable<T>` that can be `co_await`ed in a coroutine. The template arguments are the D-Bus method return types (empty for void-returning methods). The awaitable returns `void`, a single value, or a `std::tuple` for multiple return values:
+
+```c++
+        // In a coroutine context:
+        auto result = co_await concatenatorProxy->callMethodAsync("concatenate")
+                                                 .onInterface(interfaceName)
+                                                 .withArguments(numbers, separator)
+                                                 .getResultAsAwaitable<std::string>();
+        std::cout << "Got concatenate result: " << result << std::endl;
+        // If an error occurs, sdbus::Error is thrown when co_await completes
+```
+
 ### Marking client-side async methods in the IDL
 
 sdbus-c++-xml2cpp can generate C++ code for client-side async methods. We just need to annotate the method with `org.freedesktop.DBus.Method.Async`. The annotation element value must be either `client` (async on the client-side only) or `client-server` (async method on both client- and server-side):
@@ -1286,7 +1310,7 @@ sdbus-c++-xml2cpp can generate C++ code for client-side async methods. We just n
 </node>
 ```
 
-An asynchronous method can be generated as a callback-based method or `std::future`-based method. This can optionally be customized through an additional `org.freedesktop.DBus.Method.Async.ClientImpl` annotation. Its supported values are `callback` and `std::future`. The default behavior is callback-based method.
+An asynchronous method can be generated as a callback-based method, `std::future`-based method, or C++20 awaitable-based method. This can optionally be customized through an additional `org.freedesktop.DBus.Method.Async.ClientImpl` annotation. Its supported values are `callback`, `future` and `awaitable`. The default behavior is callback-based method.
 
 #### Generating callback-based async methods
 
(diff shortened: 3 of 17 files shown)


alexcani:
Hi there!
There are a few things I would appreciate your help clarifying:
- I tried my best to follow the existing code style. It would be great to have something like a .clang-format file in the project, though.
- I was not sure how to handle the copyright part. I added my name/email to new files that I've created, and left the other ones untouched. Please let me know if I should do it differently
- This is C++20, however we claim to be backward-compatible to C++17 at the public API level. This breaks it. I'd like your input on how to handle it. We could conditionally compile this based on the presence of "__cpp_impl_coroutine" feature macro.
- I left the constructor for Awaitable public to avoid needing to befriend `details::Proxy` AND `AsyncMethodInvoker`. this also removed the need for forward declarations in the file. i thought it looked good and accepted the tradeoff. i can change it if you prefer to have all "infrastructure" types constructors private.
- The integration tests are mostly a replic of the ones for async methods. To be completely transparent, i had Claude write 98% of the tests (but i did review the code myself), meaning I didn't put much effort in coming with new test cases. Let me know if there's a specific corner case you'd like to test

Also thanks for the work you put in the project, I always had a good time using it. Please let me know if there are further changes or design considerations needed

sangelovic:
Hey @alexcani. Thanks a lot for contributing and submitting the PR! I started looking into that but am a bit overloaded currently, I'll definitely come back to this asap.

sangelovic:
@alexcani 

Hi! I reviewed the PR. Appears to me to be one of the best, well-done PRs I've reviewed on this project. It is enjoyable to see changes done to nice little details and across all levels. I have no experience with coroutines, so I had to do the research first. Still, given lack of experience, I don't feel 100% confident about all the details of good coroutines design. But given the quality of your work, I think I can trust you.

However, I did some deeper reflection about things like overall design, thread coordination, exception handling, etc. -- and when I tried to think about them out-of-the-box, I got back to your solution, understanding your motives. Again, I see you've put an aspect of quality and craftsmanship to your work, and this is something I highly appreciate.

One thing I was asking myself was -- in `AwaitableData` template, could we simply use `std::variant<>` for both the result and the exception, instead of `std::optional` for one and the other? I think that this

* would idiomatically be more self-revealing that only one of those is available at a time, and thus would eliminate potential confusion of a reader about combinations resulting from having two `std::optional`s (like can it be that both are set at the same time? Or can it be that both are unset?),
* would need only one member instead of always two, making it simpler,
* while introducing no extra performance penalty.

I've already prepared a local patch for that. It works nicely. I can push that to your branch if you're okay with that, for you to review. Let me know. In the code I touched, I also changed the formatting a bit to stay consistent with the surrounding code.

Regarding code style, yes, clang-format would be great. I have that on my TODO list.

Regarding copyright, all fine with me.

Regarding API, yes, it is currently C++17 compatible. Would be great to use feature macros to have this feature conditionally available.

Regarding Awaitable public c-tor, I don't have a strong opinion, but for consistency with other similar solutions in the library, I would slightly tend towards private c-tor and befriending the relevant classes. It's already done this way in other places in the library. Additionally, I would add `assert(data_ != nullptr);` in the c-tor body to express the invariant of always non-null ptr clearly.

Tests LGTM.

Clang-tidy job is failing -- shall I address the reported issues, or will you?

I'm glad you've enjoyed using the library so far. That was my ultimate design goal, and I've put my best into it.

alexcani:
Hi!

Thanks for taking the time to review and also for the kind words!

Of course, feel free to push the patches you have prepared. std::variant
makes sense as well, i think I didnt use it just based on personal taste.

I'm currently in vacation with no access to my setup (replying to this via
email on my phone), so it will take a couple weeks before I'm able to put
some work into it, however i can work on the other points (c++17 compat,
the privatr ctor + friends, the assertions and the linter issues) when I'm
back if you prefer. Just let me know!
Also i was thinking if we need some tests for making sure the feature is
properly toggled via the feature macros on different compilers.

Kind regards,
Alex

Em qua., 25 de mar. de 2026, 8:51 PM, Stanislav Angelovič <
***@***.***> escreveu:

sangelovic:
OK, I'll commit my patch. And no problem -- if I find time then I may do the adjustments myself. If not, you can do it when you are back. Enjoy your vacation.

sangelovic:
Hi @alexcani. Today I found some time so I pushed the `std::variant` commit + did the remaining stuff we agreed upon. Would you please review the commits? (You can review commit by commit as each tries to do a self-contained logical thing.)

C++17 compat was not that easy. I tried to conditionally remove all functions using awaitables, but we have virtual methods and there we get into troubles if the library is built with C++20 and the client with C++17 (broken virtual table contract). So I kept the API even when `__cpp_lib_coroutines` is not defined, but

1. I isolated the conditional changes in the awaitable header,
2. and have the `Awaitable` class and its layout to maintain the ABI contract in all cases, but with no methods and no use of `std::coroutine_handler<>`.

Does it look OK to you?

With new major release v3 in the future, we may remove that conditional build and completely switch to C++20 or C++23 in public API as well.

alexcani:
> Hi @alexcani. Today I found some time so I pushed the `std::variant` commit + did the remaining stuff we agreed upon. Would you please review the commits? (You can review commit by commit as each tries to do a self-contained logical thing.)
> 
> C++17 compat was not that easy. I tried to conditionally remove all functions using awaitables, but we have virtual methods and there we get into troubles if the library is built with C++20 and the client with C++17 (broken virtual table contract). So I kept the API even when `__cpp_lib_coroutines` is not defined, but
> 
> 1. I isolated the conditional changes in the awaitable header,
> 2. and have the `Awaitable` class and its layout to maintain the ABI contract in all cases, but with no methods and no use of `std::coroutine_handler<>`.
> 
> Does it look OK to you?
> 
> With new major release v3 in the future, we may remove that conditional build and completely switch to C++20 or C++23 in public API as well.

Hi! Thank you for the work, I like the std::variant better as well!
I checked all the commits, and i think there might be a small issue with the C++17 compat regarding the ABI.
The `AwaitableData` class has different layout depending on the `__cpp_lib_coroutines` flag, which will cause member offsets to change in the client side vs lib side. In particular, the handle is declared between the result and the status members. This means that the `data->status.exchange` call on the `include/sdbus-c++/ConvenienceApiClasses.inl` file will potentially explode, as a C++17 client could run the code on a `AwaitableData` object created by a C++20 lib.

I did some thinkering with out AI overlord and came up with a solution by keeping the layout the same via a type-erased handle (a.k.a. a pointer) in `AwaitableData`. Here's a proposed patch.

```cpp
diff --git a/include/sdbus-c++/Awaitable.h b/include/sdbus-c++/Awaitable.h
index 2a9e638..e42742b 100644
--- a/include/sdbus-c++/Awaitable.h
+++ b/include/sdbus-c++/Awaitable.h
@@ -69,15 +69,14 @@ namespace sdbus {
     {
         using result_type = std::conditional_t<std::is_void_v<T>, std::monostate, T>;
         std::variant<result_type, std::exception_ptr> result;
-#ifdef __cpp_lib_coroutine
-        std::coroutine_handle<> handle;
-#endif // __cpp_lib_coroutine
+        void *erasedCoroutineHandle = nullptr;
         std::atomic<AwaitableState> status{AwaitableState::NotReady};
 
         void resumeCoroutine()
         {
 #ifdef __cpp_lib_coroutine
-            handle.resume();
+            if (erasedCoroutineHandle)
+                std::coroutine_handle<>::from_address(erasedCoroutineHandle).resume();
 #endif // __cpp_lib_coroutine
         }
     };
@@ -118,7 +117,7 @@ namespace sdbus {
         // resume the coroutine.
         bool await_suspend(std::coroutine_handle<> handle) noexcept
         {
-            data_->handle = handle;
+            data_->erasedCoroutineHandle = handle.address();
 
             // Attempt transition from NotReady to Waiting.
             AwaitableState expected = AwaitableState::NotReady;
```
The coroutine handles are trivially copyiable and can be easily erased and recovered via `.address()` and `from_address()`. Usage is still gated behind the flag, but at least we keep the ABI consistent.

What do you think?

sangelovic:
@alexcani Hi, thanks for the review!

I looked at your described case and this is what I see: The original idea of mine is that `AwaitableData<T>` should not be used in C++17 client code at all -- there is no support for coroutines in there, so awaitable-tagged methods shall not be used. This is not enforced in code, however; it is only documented. Not using this API in C++17 means having no ABI problem.

But still, if a C++17-based client would call the `AsyncMethodInvoker::getResultAsAwaitable()` you are mentioning as an example, the `AwaitableData` both construction and use happen in that method, meaning both get compiled into client's code, both with C++17, i.e. both with consistent `AwaitableData` layout seen there. The incosistent `AwaitableData` layout issue could potentially happen with `IProxy::callMethodAsync()`, where `AwaitableData` is created inside sdbus-c++ lib (C++20) and its members accessed by a C++17-based client. However:

1. `AwaitableData` is inaccessible for clients as it is private within `Awaitable` class, and
2. again, C++17-based clients have no interest and should not use these functions as they don't have coroutines. (Best would be to conditionally remove that API, but then we have this `IProxy` vtable inconsistency problem, which is real problem.)

Does this make sense to you, or perhaps I am missing something?

alexcani:
Hi! I agree that within reasonable use it is unlikely that a C++17 client will interact with a C++20 compiled `AwaitableData`, but to me it feels a bit brittle. We're not really warning the C++17 user that they shouldn't use coroutine-related functionality. So instead of flirting with UB I'd prefer to just keep the layout consistent. We even reduce line count by 1 in the process and have one less ifdef. The null-check for `erasedCoroutineHandle` could also be dropped since `resumeCoroutine` is only called when state is `AwaitableState::Waiting`, which implies a coroutine was set in the first place.

I don't want to enter the world of "what-if's", but it's relatively easy for the client to generate code that would trigger UB:
```cpp
// Get proxy from the library
std::unique_ptr<sdbus::IProxy> proxy = /* ... */;
// Call virtual method -> lib method is called with AwaitableData containing handle
auto aw = proxy->callMethodAsync(/*...*/, sdbus::with_awaitable);

// Since Awaitable contains 1 member, its address coincides with the member's
auto* sharedPtr = reinterpret_cast<std::shared_ptr<sdbus::AwaitableData<MethodReply>>*>(&aw);

// Memory stomping, since 'status' is in a different offset than the one seen from the client
(*sharedPtr)->status.exchange(sdbus::AwaitableState::Completed, std::memory_order_acq_rel);
```

Of course the client is shooting itself in the foot (and reinterpreting the private layout of Awaitable is already UB itself), and the code is a bit hacky and makes no practical sense, but the point is that the ABI break exists, we just lack a convenient access path to it at the moment.

It's not a deal-breaker for me as well if you prefer to keep it as-is, as well, but I think it would be worth it.

sangelovic:
In general, I would be strongly reluctant to make the code more complex for the hypothetical edge case of someone anyway determined to shoot his foot (where is the boundary of sane vs. totally safe API design then, right?), but what you propose seems acceptable -- the code is not more complex, it's "just" a bit less type-safe and less expressive.

I would have yet another idea -- simply make the coroutine_handle the last member of `AwaitableData`. This is a typical way how ABI-compatible changes are done to vtables, for example (new virtual functions being added to the end of the class, so new virtual functions don't change offsets of existing ones). It should solve your concerns, while keeping the higher type-ness and expressiveness of the code. I would personally vote for this one, but I'll also accept your proposal -- make your choice (I guess you'll push the commit?) Thanks a lot.

alexcani:
> In general, I would be strongly reluctant to make the code more complex for the hypothetical edge case of someone anyway determined to shoot his foot (where is the boundary of sane vs. totally safe API design then, right?), but what you propose seems acceptable -- the code is not more complex, it's "just" a bit less type-safe and less expressive.
> 
> I would have yet another idea -- simply make the coroutine_handle the last member of `AwaitableData`. This is a typical way how ABI-compatible changes are done to vtables, for example (new virtual functions being added to the end of the class, so new virtual functions don't change offsets of existing ones). It should solve your concerns, while keeping the higher type-ness and expressiveness of the code. I would personally vote for this one, but I'll also accept your proposal -- make your choice (I guess you'll push the commit?) Thanks a lot.

Hi! I think that's a good middle ground. I have pushed the change moving the handle last in the class.
Anything left from your side?

sangelovic:
> Hi! I think that's a good middle ground. I have pushed the change moving the handle last in the class. Anything left from your side?

OK, great, I think we are good to go. Thank you for your contribution! Very appreciated.

